PR #9, opened 2018-05-20
Add simple docker deployment

kpcyrd:
I've only lightly tested this, but it should work to get started.


dansup:
@kpcyrd Thanks!

PR #12, opened 2018-05-26
Malay translation

Ahmad-Amsyar:
Add malay translation.


dansup:
@Ahmad-Amsyar Thanks for the translations!

PR #15, opened 2018-05-26
Add German Translation

timmyRS:


dansup:
@timmyrs Thanks for the translation!

PR #20, opened 2018-05-26
profile.php Translate French

n-haddag:


dansup:
@OpenAlgeria Thanks, I don't know how profile.php slipped into that commit, I was going to push it after I added some more string but it can wait.

n-haddag:
@dansup Ok no problem

PR #21, opened 2018-05-27
notification.php Translate French

n-haddag:


dansup:
@OpenAlgeria You are fast! Thanks again for the contribution.

PR #25, opened 2018-05-27
escaping single quote mark

Grazouillis:


mkljczk:
What about `’`?

Grazouillis:
French keyboards don't have the character you suggest, but why not.
If you don't like `\'` , `&apos;` is even more relevant.

PR #34, opened 2018-05-27
Spanish traslation

yabirgb:


dansup:
@yabirgb Thanks for the translations!

PR #35, opened 2018-05-28
Greek translation

qwazix:
Added translations for greek (el). Also added DS_Store to .gitignore


dansup:
@qwazix Thanks for the translations!

PR #38, opened 2018-05-29
Workable Docker container definitions and docker-compose runtime

moritzheiber:
These are a couple of changes that are (IMHO) necessary to get Pixelfed running as an isolated Docker container:

### Dockerfile
- Pin the PHP version in the dependent Docker `FROM` target. Otherwise we might get into trouble when there are version updates
- Pin the `composer` version and install it from its official source, including a checksum verification to ensure we aren't installing malicious code
- Pre-populate a slew of needed directories inside the image (you can still use a `bind` mount to overwrite them) to ensure the Docker image is able to run without having an sources mounted into the image
- Use `www-data` as the user which owns the storage directory structure.

### docker-compose.yml

You will need to run the following two commands in order to get the stack to run:

```
$ cp .env.example .env
$ docker-compose run --rm php php artisan key:generate --show
```
Save the output somewhere and then run the following command:

```
$ APP_KEY="<the-output>" docker-compose up
```

After a short while the application should be available at [http://localhost:3000](http://localhost:3000).

I would've written a README section for this, but I was unsure what the plans were in terms of documentation.


Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+  docker-php-ext-install pdo_mysql pcntl && \
+  pecl install imagick && \
+  docker-php-ext-enable imagick pcntl imagick && \
+  curl -LsS https://getcomposer.org/download/${COMPOSER_VERSION}/composer.phar -o /tmp/composer.phar && \
+  echo "${COMPOSER_CHECKSUM}  /tmp/composer.phar" | sha256sum -c - && \
+  install -m0755 -o root -g root /tmp/composer.phar /usr/bin/composer.phar && \
+  ln -sf /usr/bin/composer.phar /usr/bin/composer && \
+  mkdir -p /var/www && \
+  install -d -m0755 -o www-data -g www-data /var/www/html/pixelfed \
+    /var/www/html/pixelfed/storage \
```
kpcyrd: We can't set those permissions at buildtime, if we mount a directory to this location it's going to be owned by root with default permissions. We might have to wrap around the php entrypoint to initialize those at runtime.

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+    /var/www/html/pixelfed/storage/logs \
+    /var/www/html/pixelfed/storage/framework/sessions \
+    /var/www/html/pixelfed/storage/framework/views \
+    /var/www/html/pixelfed/storage/framework/cache && \
+  rm /tmp/composer.phar && \
+  apk del --purge .build
 
-WORKDIR /var/www/html
-COPY . .
+COPY --chown=www-data . /var/www/html/pixelfed/
```
kpcyrd: The application should not be able to modify itself. Directories that need to be writable should be set to writable explicitly.

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+
+WORKDIR /var/www/html/pixelfed
+USER www-data
 RUN composer install --prefer-source --no-interaction
+
+VOLUME ["/var/www/html"]
```
kpcyrd: This is going to make `/var/www/html` an empty folder when used outside of docker-compose. This should point to the `storage/` folder instead.

docker-compose-running.yml
```diff
@@ -0,0 +1,64 @@
+---
```
kpcyrd: The original file was super barebones to get started, feel free to merge all of your changes into one file. :)

docker-compose.yml
```diff
@@ -3,37 +3,58 @@ version: '3'
...
     environment:
       - DB_HOST=mysql
       - DB_DATABASE=pixelfed
-      - DB_USERNAME=${DB_USERNAME}
-      - DB_PASSWORD=${DB_PASSWORD}
+      - DB_USERNAME=${DB_USERNAME:-pixelfed}
+      - DB_PASSWORD=${DB_PASSWORD:-pixelfed}
       - REDIS_HOST=redis
-      - APP_KEY=${APP_KEY}
+      - APP_KEY=${APP_KEY:-app_key}
```
kpcyrd: This key is used internally for security stuff and must not have a default value.

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+    /var/www/html/pixelfed/storage/framework/views \
+    /var/www/html/pixelfed/storage/framework/cache && \
+  rm /tmp/composer.phar && \
+  apk del --purge .build
 
-WORKDIR /var/www/html
-COPY . .
+COPY --chown=www-data . /var/www/html/pixelfed/
+
+WORKDIR /var/www/html/pixelfed
```
kpcyrd: If the webroot changes, the nginx.conf has to be edited as well.

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+USER www-data
```
kpcyrd: The dependencies are explicitly installed as root to prevent the application from editing them.

kpcyrd (changes requested):
hey, thanks for your help! There are some things that might need attention, mostly to ensure the application is harder to exploit and enforce immutability inside the container. :)

Dockerfile
moritzheiber: I'm not sure what you're implying here, could you clarify?

Dockerfile
moritzheiber: Do we know which those are?

Dockerfile
moritzheiber: No, if you're using named volumes (which should be the case for pretty much anyone considering the run Docker in a non-dev environment) the folder content is going to be a part of the volume (see the `docker-compose-running.yml` file for an example).

docker-compose-running.yml
moritzheiber: The issue here is: These files serve two different purposes. One of them has you run the containers against your development environment, the other is used to run the Docker container in a setup where the development environment is no longer required. Merging these two files wouldn't work.

docker-compose.yml
moritzheiber: It's not going to work anyway, since the app expects it to be a base64 encoded CBC-128/256 cipher. I'll remove it.

Dockerfile
moritzheiber: Maybe we can use a user different from `root` instead then? I mean, composer literally says "Don't run me as root"?

Dockerfile
moritzheiber: Yup, I was planning on submitting a second PR to include a shoe-in nginx configuration afterwards

Dockerfile
kpcyrd: `docker run -v /foo:/var/www/html/pixelfed/storage ...` is probably a common thing to be done, if `/foo` doesn't exist on the host system yet, docker is going to create it with root:root as owner and 0755 as mode. This means that the folder is empty and www-data can't write to it, unless we change the owner inside the entrypoint script. This is very common in the official docker images.

Dockerfile
kpcyrd: That should be `/var/www/html/pixelfed/storage` only, maybe @dansup can help? :)

Dockerfile
kpcyrd: We could install the dependencies as www-data and then chown all files to root afterwards. I'm pretty sure they mean "don't run composer as root on your host system" :) The risk of running composer as root in docker containers is neglectable unless you're worried about kernel 0days.

Dockerfile
kpcyrd: oh, I didn't know that. I think bind-mounting folders into the image is still a common thing to do (I'm doing this to reduce complexity of my setup).

Wouldn't making the code base part of the volume still cause problems on upgrades?

moritzheiber: None that I would be aware of. We've been using the same principle with application that need to exposed to a CGI handler ever since named volumes became "a thing".

Dockerfile
moritzheiber: Sounds good, I'll amend the PR.

Dockerfile
moritzheiber: I think it's at least the directories I'm trying to create earlier.. ?

Dockerfile
moritzheiber: Ah, now I know what you're getting at. That's true for host bind mounts, but not for named volumes. In fact, the owner/group and access permissions are vital on already created directories inside the container to make sure the volumes are getting the same set of rules assigned to them.

In essence, I think this is a problem that amounts to "host mounts" v. "named volumes" ..

moritzheiber:
@kpcyrd I tried incorporating most of your suggestions. Could you take a look again?

kpcyrd (approved):

PR #39, opened 2018-05-29
i18n Completes Occitan translation

Nouutilizaire:
So as to complete the current translation.


Quenty31:
If you like best "picture" in short, it's «fòto» instead of "fotografia".
It's up to you :)

Nouutilizaire:
Changed + new line added

dansup:
@Nouutilizaire Thanks!

PR #40, opened 2018-05-29
Danish translation done.

voss:
Danish translation finished. I HTML encoded all the Danish characters to avoid messing them up when rendered.

If anyone speaks Danish and wants to review, that would be great (:


Zatnosk:
Native danish speaker here, and I have a few comments.

**passwords.php**
- There's a typo in 'sent', "andgangskode" should be "adgangskode"

**validation.php**
- 'array' : I think "et Array" would be more correct than "en Array".
- 'json' : "JSON string" can be translated to "JSON streng".

Generally ":attribute field" is translated inconsistently; in a few places it's translated as ":attribute feltet" and everywhere else it's translated as ":attribute felt" (definite vs indefinite). I suggest ":attribute feltet" anywhere ":attribute field" is translated. Even though it's written as indefinite in english, similar indefinite usage doesn't work in danish, so a change to the definite form is necessary.

voss:
From one Jonas to another, thanks for your comments, and for catching my typos @Zatnosk  (:

You make a great point on my inconsistent translation of ":attribute field" and a change to the definite form.

On that topic, do you think it would be better to use a hyphen so that:

`:attribute felt`

becomes 

`:attribute-feltet`

Or will that affect how the text is interpreted? I'm not really a programmer, but I would suspect that `:attribute` wouldn't get interpreted the same way as `:attribute-feltet`, and would be missed altogether.

Thinking about the two different results, it would look like:

"adgangskode-feltet"

and 

"adgangskode feltet"

Wouldn't the hyphenated version be better in that situation?

Agree that "et array" is also better than "en array".

Zatnosk:
I agree that a hyphenated version would be more correct, since compound nouns are generally without spaces in Danish.

I don't know how the parser works, though, so I can tell how it would handle ":attribute-feltet". I imagine it could possibly be done with something like "{:attribute}-feltet", but again, I don't know the parser.

dansup:
@voss Is this ready to merge?

voss:
@dansup almost. Can you clarify how the parser works when it comes to adding hyphens to :attribute? The reason being, in Danish we love compound nouns, and I think it would be more correct to add a hyphen between attribute, and the Danish word for 'field'.

How do I make sure that :`attribute-field`is parsed the right way? Do I put curly braces around :attribute like so `{:attribute}-field` or is there another way to make sure hyphens won't break it?

Once you clarify, I can make the final changes and commit those, and then it's ready to merge.

dansup:
@voss Please see https://laravel.com/docs/5.6/localization#retrieving-translation-strings

voss:
@dansup thanks for the pointer. I read it, but I'm still not sure if I can append a hyphen to :attribute without putting it in curly brackets. I wasn't able to find that addressed in the documentation.

Let me use an example instead. From validation.php, my current line 32 reads:

`'boolean'              => ':attribute felt skal v&aelig;re sandt eller falsk.',`

":attribute" and "felt" should either be compounded or hyphened in Danish, so it needs to look like this:

`'boolean'              => ':attributefelt skal v&aelig;re sandt eller falsk.',` 

or 

`'boolean'              => ':attribute-felt skal v&aelig;re sandt eller falsk.',`

In both of those instances, :attribute is no longer standing alone, and so I suspect will not be parsed correct. Do I use curly braces to indicate to the parser that the :attribute part should still be replaced, like this for example:

`'boolean'              => '{:attribute}-felt skal v&aelig;re sandt eller falsk.',`

Sorry for being thick, not a dev, still learning (:

voss:
Changes ready to commit. Still need to figure out how lavarel handles compounds of :attribute and regular words. Until then the Danish translation will look a little funny to any native Danish speaker.

voss:
This PR can be merged.

PR #78, opened 2018-06-01
Prevent card from stretching on profile page

DeadSuperHero:
Still trying to understand the asset structure better; if I'm doing this wrong, let me know.

Basically I found that resizing the browser window caused the card elements to stretch in a weird way:
pixelfed

I found that setting the `height` property to `auto` on `.card-img-top` fixes this:

pixelfed

As a bonus, it doesn't disrupt any existing views.


dansup:
@DeadSuperHero Thanks!

PR #86, opened 2018-06-01
Modal for posting new posts

hellcp:
peek 2018-06-01 10-39


mkljczk:
Seems like it requires JS, but “[javascript is just a progressive enhancement in PixelFed](https://mastodon.social/@pixelfed/100099361663783498)”… Using <noscript> for previous card is a good idea…

hellcp:
@m4sk1n it's enabled through js too, so it will only change to modal mode if you got js ;)

dansup:
@hellcp Nice work! Would you mind updating this PR to change the resources/views/timeline/partial/new-form.blade.php view instead of the personal/public timeline views?

hellcp:
Will continue in other PR

Issue #63, opened 2018-06-01
"Follow" button does not change after following someone back in /accounts/activity


trwnh:
1) Go to Activity
2) Choose a notification where "x followed you"
3) Click "Follow" to follow them back

Expected behavior: Button now says "Following" and/or changes background color.
Actual behavior: Button still says "Follow" and doesn't change at all.

Error persists across refresh.

dansup:
Fixed in 2d0442b

PR #95, opened 2018-06-01
Do not show “Follow” button in notifications if already followed

mkljczk:


mkljczk:
related to #63

neilalexander:
It could be nice in this instance to show an "Unfollow" button instead of showing nothing?

mkljczk:
@neilalexander that’s to avoid common on Instagram unwanted behavior…

Issue #74, opened 2018-06-01
Own account on Discover People


Ahmad-Amsyar:
On _discover page_, sometimes I can see my own account on **_Discover People section_**, when I click follow on my own account it lead to 404 page.
I dont know how to reproduce it because it randomly occurred.

kpcyrd:
This was fixed in #108 :)

PR #108, opened 2018-06-01
Do not recommend myself

mkljczk:
Please, merge it after merging #93, I had to base my branch on it because I use sqlite on my dev instance 😃
Related to #74


dansup:
@m4sk1n Good catch, thanks!

PR #109, opened 2018-06-01
galician (gl), translating pixelfed beta

xmgz:
I've added README.md to galician folder to keep a log and for future translators

thanks


xmgz:
is there any other folder where translation files are stored? thanks

dansup:
@xmgz Nope, that is the only folder for translations! Thanks!

PR #113, opened 2018-06-01
Make navbar look better on mobile devices

### LINKED ISSUE #62: Menu looks strange on mobile (Android, FF)
screenshot_20180601-082855
The menu looks misaligned and as if it's missing text or have too wide buttons.

Android 6.0.1
FireFox 60.0.1

mkljczk:
Probably fixed #62

### DIFF
--- a/resources/views/layouts/partial/nav.blade.php
+++ b/resources/views/layouts/partial/nav.blade.php
@@ -20,13 +20,13 @@
                     <li><a class="nav-link font-weight-bold text-primary" href="{{ route('login') }}">{{ __('Login') }}</a></li>
                     <li><a class="nav-link font-weight-bold" href="{{ route('register') }}">{{ __('Register') }}</a></li>
                 @else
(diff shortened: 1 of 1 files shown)


dansup:
@m4sk1n Please use a medium or large breakpoint instead. http://getbootstrap.com/docs/4.1/utilities/spacing/#notation

dansup (changes requested):
Please use md or lg breakpoint as defined here http://getbootstrap.com/docs/4.1/utilities/spacing/#notation

hellcp:
That's not really a solution:
* You left unclosed li tag
* It's not a consistent class attatchment, which will cause same problem that it's trying to fix with any later change

Instead what would be nicer and easier to do is to add (and replace px-3 in one of) classes to nav-item elements there: px-2

So we would end up with three class="nav-item px-2" elements

hellcp:
This looks perfect <3

dansup (approved):

Issue #119, opened 2018-06-01
Admin dashboard: Username in user list


Whaxion:
Hey !
It would be great if we have possibility to see username in the user list and also a link to his profile :smiley:

Whaxion:
I just see that the link exists, but, return 404 not found

PR #122, opened 2018-06-01
Change id to username in users list and change the url

Whaxion:
All in the title, solve my issue #119


andriussev (changes requested):
1. Why not leave the id column?

2. This introduces more complexity to number of database calls since the profile load will happen for each user separately.
This brings it to a total of a `O(1+3*n)` database calls where `n` is the number of users, on this page.

Whaxion:
1. If I don't leave the id column, it's because on mobile, the table was too large.

2. Solved.

Issue #117, opened 2018-06-01
image descriptions for accessibility


ghost:
Mastodon has a feature where you can (along with the text of the post) add an image description that is added as an alt text, for screenreaders etc. Could this be a possibility in this project?

dansup:
Added in #410

PR #124, opened 2018-06-01
Add description for visually impaired field

mkljczk:
It does not federate, so a bit work-in-progress. I don't know php and activitypub enough to take care of this…
#117 
alt

### DIFF
--- a/app/Http/Controllers/StatusController.php
+++ b/app/Http/Controllers/StatusController.php
@@ -30,9 +30,9 @@ class StatusController extends Controller
       $user = Auth::user();
 
       $this->validate($request, [
-        'photo'     => 'required|image|max:15000',
-        'caption'   => 'string|max:150',
-        'nullable'  => 'string|max:420'
+        'photo'       => 'required|image|max:15000',
+        'caption'     => 'string|max:150',
+        'description' => 'nullable|string|max:420'
       ]);
 
       $monthHash = hash('sha1', date('Y') . date('m'));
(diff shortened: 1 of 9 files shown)


andriussev:
Pardon my ignorance, but considering that it is an image sharing platform, how does supporting visually impaired people help the cause?

Apart from that, it's unclear how descriptions should be handled.

Please fix the validation rule: https://github.com/dansup/pixelfed/pull/124/files#diff-cd8487f20fdf44ab25f3701e48fab225R35

ghost:
@andriussev This is also a federated platform, and presumably images will be shared outside of the initial platform, off pixelfed and onto text-based platforms like Mastodon. In addition, some people are only partially visually impaired and not completely blind (see: literally everyone who needs visual aids like glasses) and it may be helpful for them to have text that can be read via screen reader even though they can still partially see the image being shared. Also sometimes images are just busy and it's hard to see what's going on, but text might help clarify.

ghost:
Also, accessibility always helps a cause.

ghost:
Hey, a friend said 
>"i think masto federates image descriptions by including them in the name field of the Attachment"

Hopefully that's helpful!!

vilbi:
As I als user understand the federated concept, the image discription is realized via the alt image text like described here https://en.wikipedia.org/wiki/Alt_attribute?wprov=sfla1
As a quick fix, the post text can be placed there.

mkljczk:
@vilbi, that’s what the PR is doing right now

dansup:
@m4sk1n I think this is an important accessibility requirement, I will work on updating this merge request!

vilbi:
Thanks!

dansup:
Added in #410, the main difference between this PR and #410 is that its possible to set a caption for each photo instead of a single caption for every photo.

PR #125, opened 2018-06-01
Extract database calls from status show blade file

andriussev:
I've removed all database calls from `status/show.blade.php`. All of those calls are now done in `StatusController`. instead.

This makes it easier to cache the status much easier if that is ever needed.


dansup:
@andriussev Yeah, eventually I will use Redis instead of database queries if available.

Issue #121, opened 2018-06-01
Caption limit hit, no warning, cleared what I tried to post/write


mal0ki:
"The caption may not be greater than 150 characters" Which I wasn't warned about until after I had tried posting a bigger one. And everything was cleared when I tried to post.

Suggestion:
* Character counter / countdown
* Don't clear all the submitted data, ie, save the draft I guess?

Bonus:
* Possibly more than 150 characters, even if the focus is on photos, sometimes we have a lot to say about a work of art, or a photo or something.

![image](https://user-images.githubusercontent.com/28316305/40860005-049dc168-65e4-11e8-9354-8cbb63c942c2.png)

trwnh:
related to #59 (which is for bios instead of captions)

PR #127, opened 2018-06-01
Add character limit info

mkljczk:
#121


andriussev (changes requested):
Will be a pain in the ass if the validation changes. Should be moved to the configuration and parsed from there to both the message and the validation rule.

Also, should return old data if limit failed (as should other fields). Limits should also be enforced with HTML.

dansup:
@m4sk1n I will be adding a character limit in the pixelfed config so that admins can set their own limit. It will require some additional logic in the view, controller and configs

dansup:
@m4sk1n @andriussev This was merged in #150 with configurable limit.

PR #168, opened 2018-06-03
Make form a modal with js and squarify discover

hellcp:


dansup:
@hellcp Can you pull in the latest commits so I can merge this?

hellcp:
Done, replaced the icon :D

PR #176, opened 2018-06-03
Don't use JS for navbar so much in mobile view

### LINKED ISSUE #174: navbar position fixed
It would be nice to see navbar when we scroll and then apply a fixed position to this in the CSS.
(Not very significant at this time)

hellcp:
And fix #174


hellcp:
screenshot from 2018-06-03 15-50-34
screenshot from 2018-06-03 15-50-48
screenshot from 2018-06-03 16-33-03

dansup:
@hellcp Great work!

PR #180, opened 2018-06-03
New new user profile look

hellcp:
screenshot from 2018-06-03 17-37-12


dansup:
@hellcp Nice work! You are on a roll with all these UI improvements.

PR #220, opened 2018-06-06
suggestion

n-haddag:
name instead of user In the header


dansup:
@OpenAlgeria Hi, as I mentioned in IRC, I don't this is appropriate because the name is not unique and constant like the username is.

PR #222, opened 2018-06-06
Add animation for likes

hellcp:


dansup:
@hellcp I'd like to get #226 implemented before I accept animations, so this will have to wait a few days!

hellcp:
Great!

PR #230, opened 2018-06-07
Add .editorconfig

smorimoto:
[EditorConfig](https://editorconfig.org)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


hellcp:
In HTML and CSS files we got indent of 2 spaces ;)

smorimoto:
Oh! I see.
I found some css and scss files. We also have js files (It has 2 indent and 4 indent). But where is the html file?

dansup:
@imbsky @hellcp It should be 4 spaces. Thanks!

smorimoto:
@dansup Oh, not at all!

PR #236, opened 2018-06-07
WIP: Add input maxlength to prevent typing too much

trwnh:
If a user types too much and tries to submit, their entire caption will be lost.
This prevents the user from typing too much.
Future PR for JS users should show a live character count.

- [x] Caption maxlength
- [ ] Comment maxlength

I'm not even sure where comment max is defined, or what it even is...
in my testing, I can only comment up to 255 characters, but
CommentController.php has a line  'comment' => 'required|string|max:500'


trwnh:
PR is now stale after #346

PR #237, opened 2018-06-07
Footer reorganization

trwnh:
- Clicking pixelfed.org should load pixelfed.org
- Order of footer should more closely match site routes menu


hellcp:
I would change it to "Powered by PixelFed", it would be nice towards later instances to not brand them as belonging to "PixelFed.org", instead just create indicator that it's powered by this project ;)

trwnh:
Good idea

PR #281, opened 2018-06-16
Create an Apache based Docker image ready for production

kaiyou:
This PR provides an alternate ``Dockerfile`` built on top of ``php-apache``. It is a bit heavier than the fpm image but provides full support of Pixelfed in a single image. Although not perfect for large scale production, it should fit most Pixelfed setups up to a couple hundred users.

It is possible to separate the horizon worker from the main container or embed it as a separate process in the same container. This requires the use of Redis.

I also took the liberty to suggest a new ``docker-compose.yml`` based on this image, since it runs fine in our testing environment and we are currently stressing it internally before going public and enabling federation on the instance.

For those willing to test the image directly, it is available with automated builds from our branch at: https://hub.docker.com/r/tedomum/pixelfed/ (some builds are still in progress when writing these lines).


kaiyou:
Any news about this?

kaiyou:
Any news about this? The imminent upgrade will require that Docker users are able to pull the latest version but I am not sure this Dockerfile will implement all requirements since the last commit.

kpcyrd:
I'm not sure how much I can help here due to my limited experience of docker+php, maybe @moritzheiber can help.

moritzheiber:
_Happy to review this, feel free to assign me_

There are two things that strike me as odd:

1. There are two separate processes running inside a single container, which shouldn't be happening
2. Neither the entrypoint nor the start script are taking care of process handling ultimately leading to zombie processes because of a lack of child reaping

Also having all the environment variables inside the Dockerfile is a smell I wouldn't want to commit to. Maybe we can also provide an example `.env` file? Or re-locate them inside the `docker-compose.yml`.

In any case I don't really feel like your variant is any more "production ready" as the previous setup, but that's a personal opinion.

kaiyou:
I am sorry I did not see your edit at first.

Regarding the multiple processes running in a single container, this is not the only way to run the image. Actually, the provided example compose file has instructions about running horizon in a separate container. The default behavior is just set so that most users can run a single container, which is usually expected when running a single application.

Also, taking care of process handling would indeed help properly stop containers, it does not however lead to any zombie as is.

Finally, I believe a single Docker image should be self-contained, so we need a default value for environment variables that are not set by the application itself. These are set in the Dockerfile, they can very well be overridden in the compose file or anywhere, but the image can still be used and run properly in any other environment, including Kubernetes.

dansup (approved):

PR #298, opened 2018-06-18
general: Add detailed README.md

BrainBlasted:
Previously the README.md was sparse, including no intructions on setting
up PixelFed. This could be progibitive for attracting new contributors.
I have updated the README and based the install instructions on
https://gist.github.com/dansup/999de4577f864d1822d226b9ddd62ca0.

Closes https://github.com/dansup/pixelfed/issues/18


trwnh:
Why would MariaDB not be supported when MySQL is? MariaDB is a drop-in substitute for MySQL.

PR #301, opened 2018-06-18
Make nav sizing consistent

### LINKED ISSUE #300: The action bar is now way too big
The old size was great, it fit all the icons snugly, now it's wasting tons of space

trwnh:
- Logo same height as name, without extraneous padding
- Only apply padding to icons above certain size, to prevent overflow

Before | After
--- | ---
image | image

Fix #300


trwnh:
Stale PR

PR #302, opened 2018-06-19
Replace js based menu with the dashboard for non-js users

hellcp:
screenshot from 2018-06-19 03-57-44
screenshot from 2018-06-19 03-58-02


hellcp:
screenshot from 2018-06-19 04-06-28
screenshot from 2018-06-19 04-06-47

dansup:
@hellcp Non-js users will have to use the Lite UI which doesn't use the same bootstrap dropdown menu.

PR #308, opened 2018-06-20
Change heart color from magenta to reddish

trwnh:
Before | After
--- | ---
image | image

I think it looks better. The magenta was a bit weird.


trwnh:
Stale with #349

Issue #366, opened 2018-08-13
README instructions: mysql should be >= 5.7


jeroensmeets:
Running `php artisan migrate` (or migrate:fresh), I get an error for the SQL that creates the `statuses` table:

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'json null, `created_at` timestamp null, `updated_at` timestamp null) default cha' at line 1 (SQL: create table `statuses` (`id` bigint unsigned not null auto_increment primary key, `uri` varchar(191) null, `caption` varchar(191) null, `rendered` text null, `profile_id` bigint unsigned null, `in_reply_to_id` bigint unsigned null, `reblog_of_id` bigint unsigned null, `url` varchar(191) null, `is_nsfw` tinyint(1) not null default '0', `visibility` enum('public', 'unlisted', 'private', 'direct') not null default 'public', `reply` tinyint(1) not null default '0', `likes_count` bigint unsigned not null default '0', `reblogs_count` bigint unsigned not null default '0', `language` varchar(191) null, `conversation_id` bigint unsigned null, `local` tinyint(1) not null default '1', `application_id` bigint unsigned null, `in_reply_to_profile_id` bigint unsigned null, `entities` json null, `created_at` timestamp null, `updated_at` timestamp null) default character set utf8mb4 collate 'utf8mb4_unicode_ci')
```

jeroensmeets:
looks like

`entities` json null

is the cause, mysql should be version >= 5.7

dansup:
Fixed in  f718efc

PR #368, opened 2018-08-13
Add MySQL version requirement to README

jeroensmeets:
Hi dansup, as I left issue #366  I thought it would be better to create a pull request. Added the 5.7 requirement for MySQL to the readme.


mkody:
Does this means >= 5.7 **isn't** supported?

I would replace that comma with a period.

jeroensmeets:
@mkody thx for the idea, rewrote the line.

dansup (approved):

PR #383, opened 2018-08-15
Update (DM)

Quenty31:


mkljczk:
If you don't know how to use git and you want to use online editor, I can suggest you that after you make one change, go to the newly created branch (in this case https://github.com/Quenty31/pixelfed/tree/patch-1) and make next changes from here

dansup (approved):

Quenty31:
@m4sk1n Yes I usually do so, but somehow I don't know what I did :s 
Sorry!

PR #409, opened 2018-08-26
Allow new posts without captions

### LINKED ISSUE #110: Cannot post new photo with empty caption
Posting with no caption returns the error: `The caption must be a string.`

ellotheth:
Fixes #110

- [x] Posting a new item
- [ ] Atom feed item titles
- [ ] Alt text on carousel items (multi-photo statuses)
- [ ] OpenGraph description
- [ ] Comment sidebar

Would it make sense to use some sort of default caption for accessibility and metadata, _Photo by X on Y_ or something?

### DIFF
--- a/app/Http/Controllers/StatusController.php
+++ b/app/Http/Controllers/StatusController.php
@@ -69,7 +69,7 @@ class StatusController extends Controller
 
         $this->validate($request, [
           'photo.*'   => 'required|mimes:jpeg,png,gif|max:' . config('pixelfed.max_photo_size'),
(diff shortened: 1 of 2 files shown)


dansup:
@ellotheth Hello. I'm not sure if we should allow empty captions. Will need to think about this.

trwnh:
Instagram lets you have empty captions, and I don't see why we would disallow that, even if it's just `""`. In cases where the user can't think of a caption, or cases where the caption hypothetically fails to upload for some reason, the status should still post, right?

PR #425, opened 2018-08-31
Profile thumb background-image replaced by img

Simounet:
Because this is not decoration but content (and because I'll open an issue about pixel ratio).
I had some issues with npm and not sure about the right dependencies versions so let me know.
I used `object-fit` which is Edge 16+.
Have a nice day.


dansup:
@Simounet Pixelfed uses a webpack wrapper called Mix, please follow [these instructions](https://laravel.com/docs/5.6/mix#running-mix) to compile the assets.

Simounet:
I did it and rebased my branch.

Simounet:
Branch rebased and now pointing to `dev`.

dansup:
Profiles are now loaded via a vue component which uses ``````

PR #552, opened 2018-11-11
Drop privileges when starting artisan or horizon

kaiyou:


dereckson (commented):
Why gosu instead of su -u ?

dansup (approved):

PR #553, opened 2018-11-12
fixed a bash scripting bug in start.sh

koehn:
that prevented horizon from ever starting.


contrib/docker/start.sh
```diff
@@ -12,7 +12,7 @@ php artisan migrate --force
 php artisan update
 
 # Run a worker if it is set as embedded
-if [ $HORIZON_EMBED = true ]; then
+if [ "$HORIZON_EMBED" -eq "true" ]; then
```
dereckson: -eq is for integers, = is well the comparison operator to use for strings.

The real bug was the lack of "" (and that should be fixed).

dereckson (changes requested):
Keep = operator, add quotes.

dansup (changes requested):
Fix merge conflict

koehn:
Changes merged.

dansup (approved):

PR #572, opened 2018-11-20
Offer a setup hint for `composer install`

dereckson:
If the application is run without the dependencies installation step,
offer a setup hint for the forgotten step instead of die for a missing file.


dereckson:
It seems a best practice to offer informational reporting, without assumption the person installing the application can link autoload.php to Composer.

See for example https://github.com/wikimedia/mediawiki/blob/master/includes/PHPVersionCheck.php#L169 for an implementation example and https://forum.omeka.org/t/failed-opening-required-vendor-autoload-php/2725 for evidence this is not an universal knowledge.

dansup:
This would involve checking for the autoload file exists on every page load, its not worth the performance penalty.

PR #663, opened 2018-12-21
Use saner way to change cap colour

hellcp:


dansup (changes requested):
Can you compile the production assets? ```npm run production``` for minified assets

dansup:
@hellcp After more consideration I do not think this will work with the Micro UI, we can't customize bootstrap for a specific UI.

PR #676, opened 2018-12-25
[274] Add comment button to comment form where browser has 'Touch' event enabled

### LINKED ISSUE #274: Add a 'send comment' or 'submit comment' button
When entering a comment on e.g. mobile devices with touch display
- users expect a send button on the right side of the text input window
- of course with virtual keyboard as-is with 'enter' is working, but not the usual UI ;-)

cmho:
Closes #274.

Adds a check to components.js to determine if the client brower can fire "touch" events; if so, adds the .touch class to the body and displays a blue button (btn-primary class) with appropriate spacing/padding at the end of the comment input.

There might be a better way to do this using Vue? I considered using a v-if on the <input /> for the submit, but since the method I used to detect touch devices relies on using a try/catch statement, I didn't think that would work.


dansup (approved):

dansup:
@cmho Great work, thanks!

PR #701, opened 2018-12-27
Add required PHP extensions to composer.json

mattsches:
Would it make sense to add the required PHP extensions to the main composer.json? I think so. If you agree, feel free to merge this PR, otherwise just close it ;)


composer.json
```diff
@@ -6,6 +6,13 @@
     "type": "project",
     "require": {
         "php": "^7.1.3",
+        "ext-bcmath": "*",
+        "ext-ctype": "*",
+        "ext-curl": "*",
+        "ext-json": "*",
+        "ext-mbstring": "*",
+        "ext-openssl": "*",
+        "ext-redis": "*",
```
dansup: Please remove this. It will conflict with the default redis library pixelfed uses.

```suggestion
```

dansup (changes requested):

mattsches:
Used the Github GUI to "commit suggestion", not sure what's supposed to happen next ;)

dansup (approved):

PR #729, opened 2019-01-02
[700] output image descriptions for accessibility

### LINKED ISSUE #700: Image Descriptions Don't Do Anything?
Checking via both Mastodon and Pixelfed itself, making a post with an accessible image description, either via the experimental UI or post editing UI, the image description does not actually show up anywhere. Inspecting accessibility properties in Firefox (for ubuntu, 64.0) on either mastodon or pixelfed shows nothing, and the mastodon version usually would have its `title` set to the image description, exposing it on hover.

https://pixelfed.social/p/Fidgetcetera/25511 See this post for an example, there is an image description which shows up when we edit the post (although, there is apparently a character limit which does not exist in the experimental UI), but it is not exposed anywhere on the page in a way where a screen reader would actually read it.

cmho:
Closes #700.

Exposes image descriptions in alt tags, as preferred by WCAG accessibility standards. Also adds description fields to the default compose screen. Also added a caption field to the edit screen to edit the overall caption.

(take 2, fixed a few things now that my dev environment is working better.)


trwnh (changes requested):
Please resolve merge conflicts; also, in the future, PRs should target the `frontend-ui-refactor` branch instead of the `dev` branch (we're working on documenting this for contributors soon!).

carlydotwebsite:
@trwnh Fixed merge conflicts! Give it a look.

trwnh (approved):

ZoeBijl:
Why was this closed without comment? There’s still no easy way to add alt text—this is unacceptable from an image sharing platform. In addition, the alt attribute is currently set to the image’s parent div. Unfortunately that’s not how [alt attributes](https://www.w3.org/TR/html/semantics-embedded-content.html#alt-text) work.

trwnh:
@ZoeBijl I think this particular PR got stale, and since then alt text output was added in a separate PR (incorrectly, as you point out). Also, adding alt text is a separate issue and has been possible for quite some time, though made easier with updates.

For the former, I'll open an issue for applying the alt attribute to the img and not to the div. For the latter, there's not much that can be done to make it easier except perhaps the media toolbar should be shown by default.

Issue #723, opened 2018-12-30
Cannot upload avatar for the 1st time


loicm:
On a local fresh Pixelfed install with only one user (me).

When I want to change my avatar it does not work and I have a "404 error page".

Investigating the code from the `AvatarController::store()` method I think I found why.

See: https://github.com/pixelfed/pixelfed/blob/dev/app/Http/Controllers/AvatarController.php#L36

As I have no avatar yet the `firstOrFail()` raise the 404.

- If I should have had an Avatar linked to my profile, the error is elsewhere at account creation
- I suggest to replace `firstOrFail()` by `firstOrNew([])`. So if there is no Avatar a new one is created. I can provide a PR if needed. I have tested the solution on my instance and it's working.

Findus23:
That sounds pretty much like the solution for #654 (at least I also had the same error when setting it for the first time)

loicm:
Yes it is. I should have had linked to this issue,thanks for having done it 😊

dansup:
Duplicate of #654

loicm:
I can understand the duplicate, but what about my solution?

Issue #654, opened 2018-12-18
Can't change profile picture on pixelfed.social


v7rzlhb3xg:
Steps to reproduce:

1. Go to Account Settings
2. Click "Change Profile Photo"
3. Choose file, "Upload" button remains dimmed but clickable
4. Click "Upload," nothing happens.

dansup:
@dar-irl There is a file size limit on avatars of [2mb](https://github.com/pixelfed/pixelfed/blob/dev/app/Http/Controllers/AvatarController.php#L22). I will update that to 8mb and add a file size limit warning.

v7rzlhb3xg:
The picture I'm trying to upload is still only 70kb, which wouldn't be enough to even hit the original 2mb limit though.

pgpattison:
I'm still having this same problem. Bug #637

pgpattison:
Nope. Still doesn't work. This is clearly not a file size issue. All of the images I've tried to upload as my avatar are smaller than 100 KB.

dansup:
@pgpattison Can you upload the image to this issue please?

pgpattison:
![blog_portrait](https://user-images.githubusercontent.com/16155970/50389374-d8241980-06e6-11e9-8e89-224caf8839be.png)
![my-libravatar](https://user-images.githubusercontent.com/16155970/50389375-d8241980-06e6-11e9-9c8a-302995623424.jpg)
![social_photo_2016_400x400](https://user-images.githubusercontent.com/16155970/50389376-d8241980-06e6-11e9-96ff-a56cb810f485.jpg)

Tried and failed with all three of these images. Also tried and failed with several other arbitrary images. "Upload" button never enables after selecting file, regardless of file size, extension, image dimensions, etc.

benoitjpnet:
Having the same issue (I guess?) with my personal instance.

![screenshot_20181227_212734](https://user-images.githubusercontent.com/1057913/50493521-61f12100-0a1e-11e9-9b48-b474db9ef33c.png)

Findus23:
Same on my instance.
The exact stack strace is 

<details>

```json
{
    "message": "No query results for model [App\\Avatar].",
    "exception": "Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException",
    "file": "/var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Foundation/Exceptions/Handler.php",
    "line": 200,
    "trace": [
        {
[...]

PR #747, opened 2019-01-06
Fix profile picture update

loicm:
Related to issue #654 , see also issue #723

It looks like User does not have an Avatar record at first. So getting the Avatar with `firstOrFail()` raise a 404 error page.

I suggest we can get the Avatar and if it does not exist, creating an instance on the fly with `firstOrNew()`. So that everything works in both contexts.


todoneunl:
This gives this error:

 Symfony \ Component \ Debug \ Exception \ FatalThrowableError (E_RECOVERABLE_ERROR)
Too few arguments to function Illuminate\Database\Eloquent\Builder::firstOrNew(), 0 passed in /var/www/pixelfed/app/Http/Controllers/AvatarController.php on line 36 and at least 1 expected

PR #748, opened 2019-01-06
manifest.json added for webapp declaration

Simounet:
Hi there,
With this manifest, on mobile, you can add PixelFed to your launcher from your browser and use it as a standalone application. It's the first step for PF as a [PWA](https://developer.mozilla.org/en-US/docs/Web/Apps/Progressive). Next step, service worker with offline handling.
Have a nice day.


trwnh (approved):
Only thing that strikes me as odd is the 153x152 favicon.png, but that can be dealt with in a separate PR.

Simounet:
I agree, this will be my next small PR if you want. :p

dansup (approved):

PR #757, opened 2019-01-09
[744] Add scripting to trigger like button on post double-click

### LINKED ISSUE #744: Double click to like
I'd be nice if you could simply double click a post to like it. This is the default behaviour in IG and lots of users are used to it.

Not sure if this has been already asked or it's planned

Thanks!

cmho:
Closes #744.

On double-click event from timeline post and individual post pages, status will receive "Like" activity.
Also removes a comma in PublicApiController.php that was causing posts to not load? I can split that out if need be.


public/js/activity.js
```diff
@@ -1 +1,92 @@
...
\ No newline at end of file
+/******/ (function(modules) { // webpackBootstrap
```
dansup: @cmho Great PR, can you remove the compiled assets?

dansup (changes requested):

dansup (approved):

PR #766, opened 2019-01-10
[756] Change post interaction h3s to buttons

### LINKED ISSUE #756: Change post interaction buttons to <button> element for accessibility
Right now they're marked up as h3 tags, which can't be tabbed to or clicked with keyboard navigation. If I recall correctly, the best practice for making this kind of functionality accessible is to use the button tag, which I think should be an easy-ish switch?

cmho:
Closes #756.

Swaps h3s in post interaction/reaction links to button elements for accessibility purposes, and adds some slight styling to remove unnecessary styling from the buttons.


dansup (changes requested):
Please remove compiled assets

carlydotwebsite:
@dansup done!

PR #768, opened 2019-01-10
Make NSFW shade the same size as its container

trwnh:
Before | After
--- | ---
image | image

https://pixelfed.social/p/trwnh/30939

Potential next step would be to generate blurred thumbnails server-side and apply them as `background-image`.

### DIFF
--- a/resources/assets/sass/custom.scss
+++ b/resources/assets/sass/custom.scss
@@ -306,11 +306,19 @@ details summary::-webkit-details-marker {
 }
 
 .details-animated > summary {
-  display: block;
+  display: flex;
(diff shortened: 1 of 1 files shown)


hellcp (approved):
I feel like hardcoding height is not the best design, but it's done like that for presenter already, would be nicer to have a variable for that though, but that is something we can tackle at another time, LGTM

trwnh:
Yeah, I just applied the same min-height rule that's currently being applied to the container. A more comprehensive refactor is a bit too much for right now

dansup (approved):

PR #777, opened 2019-01-11
Test on PHP 7.3

DanielRuf:


dansup:
@DanielRuf PHP 7.3 is not supported yet.

DanielRuf:
Why not test it then and add it to the allowed failures?

DanielRuf:
Regarding https://github.com/pixelfed/pixelfed/blob/dev/composer.json, it supports PHP 7.3.0, it also supports PHP < 8.0.0. (^)

PR #793, opened 2019-01-23
[730] Do image filtering via imagemagick

### LINKED ISSUE #730: Filter effects do not federate to Mastodon
If you choose a filter effect to be applied to pictures, the picture itself is federated to Mastodon, but no effect is applied.

Original: https://pixelfed.social/p/renatolond/28081
Federated: https://masto.donte.com.br/@renatolond/101346251505180944

cmho:
Closes #730.

Added a class to implement filter effects as part of images rather than implemented as CSS filters which leverages imagemagick to replicate the filtering effects and added it to the ImageUpdate job. I also added a column in the media table to keep a separate copy of the original image, unprocessed, in case the user goes back and changes the filter later.

I'm still using CSS effects in the compose window because the filter code doesn't run that fast. Eventually we'd want to remove the filter classes from the timeline, maybe, but maybe include some kind of flag to show it only if the image hasn't been processed yet?

### DIFF
--- a/app/Http/Controllers/InternalApiController.php
+++ b/app/Http/Controllers/InternalApiController.php
@@ -13,14 +13,14 @@ use App\{
     Profile,
     StatusHashtag,
     Status,
-    UserFilter,
+    UserFilter
 };
 use Auth,Cache;
 use Carbon\Carbon;
 use League\Fractal;
 use App\Transformer\Api\{
     AccountTransformer,
-    StatusTransformer,
+    StatusTransformer
 };
 use App\Jobs\StatusPipeline\NewStatusPipeline;
 use League\Fractal\Serializer\ArraySerializer;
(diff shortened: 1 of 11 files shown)


hellcp:
I love this, offsetting processing for later is an amazing idea. Ideally, we could have the image posted in the original version until the job has been finished, with class and a spinner in the corner, then display the new image in the interface without the class and spinner.

dansup:
@cmho A few issues with this commit:
- GD is the default driver for image optimizations
- I don't like it being in the ImageUpdate job, it should be its own job after it is optimized
- I am planning on using [rustagram](https://github.com/ha-shine/rustagram) because its much faster, and the filters are more accurate to the instagram.css filters.

screen shot 2019-01-23 at 1 10 03 pm

You can see how different it is, left was generated with rustagram and right with this PR.

carlydotwebsite:
Fair! I was trying to keep to PHP solutions, but I know Rust is really speedy, so I'm sure you're right that it'll work better.

PR #804, opened 2019-01-31
mariadb works from 10.2.7 onward

rhonda:
I assume the not supported status of MariaDB was due to missing JSON Data Type. That got introduced in 10.2.7 according to the docs[1] so it might make sense to update the README accordingly.

[1] https://mariadb.com/kb/en/library/json-data-type/

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ testing and development.
 
 ## Requirements
  - PHP >= 7.1.3 < 7.3 (7.2.x recommended for stable version)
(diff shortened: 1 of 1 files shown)


dansup:
@rhonda Hi, the readme has been refactored and no longer lists the requirements. The requirements will be moved to the documentation!

rhonda:
@dansup When moving it it would had been nice to apply the patch to the documentation then. It still says there that mariadb isn't supported.

MrPetovan:
Thanks for the hint!

PR #815, opened 2019-02-06
Attempt to add image to RSS feed

### LINKED ISSUE #809: Add Image URL to Atom feed
I see that you added Atom support (#37), but it would be super handy to add the image URL itself (or at least a thumbnail) to the feed. This would allow higher quality posting i.e. to Twitter via IFTTT, drawing more people to Pixelfed.

cassidyjames:
Not 100% sure this is right, but giving it a try. Should fix #809


dansup:
@cassidyjames I think this may be a duplicate of #813. I noticed the XML is a bit different in #813 and this PR so I will test them before I merge.

cassidyjames:
@dansup yeah I saw that after I'd investigated a bit and decided on an approach. If #813 is better or more comprehensive, then that makes sense!

leuc:
Atom and RSS are two different standards with different namespaces. Most if not all Feed readers support both standards.

The Atom namespace does not provide any element to include images, so the common solutions is HTML with img tag inside the summary tag. This is what #813 does.
 
RSS support would have to been added, currently it's only a Atom Feed. They can not be mixed together.

aarontrom:
@dansup Do you plan to add RSS as well or keep only the Atom Feed for now?

PR #858, opened 2019-02-17
Fix stacking icon from being on top of the header

Simounet:
Hi there,
If you scroll with a post containing more than one pic, you can see the stacking icon being on top of the header. This commit fix that.
Have a nice day.


trwnh (approved):

dansup:
@Simounet Thanks, its fixed in latest commit!

PR #904, opened 2019-02-28
Uses hashtag slug to build link

### LINKED ISSUE #40: Wrong hashtag link in post view
The hashtag link in the post view ([here](https://photos.homecomputing.fr/p/sanpi/115) for example) uses the hashtag name instead of the slush to build the link. Therefore with an uppercase hashtag the link lead to a 404 page.

### LINKED ISSUE #895: Wrong hashtag link in post view
The hashtag link in the post view ([here](https://photos.homecomputing.fr/p/sanpi/115) for example) uses the hashtag name instead of the slush to build the link. Therefore with an uppercase hashtag the link lead to a 404 page.

sanpii:
Fixes #895 

The query to fix existing links:
```sql
BEGIN;

CREATE FUNCTION lower_hashtag(rendered varchar) RETURNS varchar
AS $$
DECLARE
    match text;
BEGIN
  FOR match IN (select (REGEXP_MATCHES(rendered, '/tags/(.+)?src=hash', 'g'))[1])
  LOOP
    rendered = replace(rendered, match, lower(match));
  END LOOP;

  RETURN rendered;
END;
$$  LANGUAGE plpgsql;

UPDATE statuses SET rendered = lower_hashtag(rendered);

DROP FUNCTION lower_hashtag;

COMMIT;
```

### DIFF
--- a/app/Util/Lexer/Autolink.php
+++ b/app/Util/Lexer/Autolink.php
@@ -676,7 +676,7 @@ class Autolink extends Regex
         $hash = StringUtils::substr($tweet, $entity['indices'][0], 1);
         $linkText = $hash.$entity['hashtag'];
 
-        $attributes['href'] = $this->url_base_hash.$entity['hashtag'].'?src=hash';
+        $attributes['href'] = $this->url_base_hash.str_slug($entity['hashtag']).'?src=hash';
         $attributes['title'] = '#'.$entity['hashtag'];
         if (!empty($this->class_hash)) {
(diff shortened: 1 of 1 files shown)


dansup:
@sanpii We are already using ```str_slug()``` and you need to pass false as the 3rd argument or else it converts the slug to english!

https://github.com/pixelfed/pixelfed/blob/dev/app/Jobs/StatusPipeline/StatusEntityLexer.php#L93

PR #959, opened 2019-03-05
Change zh-tw to zh-TW

Aditoo17:
Changed the `zh-tw` folder to `zh-TW` because it's a standard and it will probably make the translation appear in `/site/languages`


dansup:
@Aditoo17 The reason its not appearing in ```/site/languages``` is because the language list is cached for 14 days. I reduced it to 1 day which will fix this!

Issue #765, opened 2019-01-10
Import from Instagram


madranet:
I don't want to sound overly-critical. I'm all in favour of Open-Sourcery and providing alternatives to the big data-slurping websites out there. So I was excited to learn about Pixelfed. But...

That excitement lasted for as long as it took me to realise there's no way to import existing photos from Instagram.  

It never fails to amaze me that so many developers seem to think that simply providing a more ethical alternative to the privacy-raping big boys out there is sufficient incentive to get people to switch to your new product. Even if doing so means they need to abandon all the time and effort they've put into building up a 'portfolio' / 'archive' / 'following' / 'history' [call it what you will] on the sites you're hoping to replace. Almost all such open source projects make a point of emphasising _"no lock in"_ and assuring you you can get your data out at any time. Yet none of  them seem to provide any strategy for importing your existing data from rival sites, in the first place.

And yet, this keeps on happening:

 App.net, Diaspora, Quitter... [to name but three that I've used in the past] ...all offering open-source or ad-free alternatives to the big players. But none of them considering it important to provide any way for new users to bring their _"stuff"_ with them from the existing sites. And all three projects subsequently dead or dying from lack of interest and users.  [and I fully expect the current poster boy for open source Social Networking, Mastodon, to go the same way, for the same reason].

Seriously devs. I know this isn't a specific issue _per se_. But you really need to consider this as a top priority, if Pixelfed is to have a snowball's chance in hell  of gaining any traction at all. 

Your product might be the most ethical, most open-source, most libertarian project in the world. But Instagram have several years and several hundred million users' head start on you. Expecting people to just abandon everything they've put into their Instagram accounts over the years and flock to Pixelfed instead is both incredibly naive and borders on insulting. It's effectively saying to them that you consider all that prior work they've done to be worthless.

Just my well-meaning tuppence worth. Feel free to close and mark as irrelevant, if you wish.

dansup:
@madranet IG import has been on the roadmap for awhile, https://mastodon.social/@pixelfed/100087667113289786

jaredmoody:
What does "on the roadmap" mean? I don't see an import function on pixelfed.social and I don't see an open issue to implement this, so I'm not sure. Was an import feature completed? Still to be completed?

Thanks!

trwnh:
[...]

PR #1046, opened 2019-03-18
Instagram Import

dansup:


app/Http/Controllers/Import/Instagram.php
```diff
@@ -21,6 +22,13 @@ public function instagram()
 
     public function instagramStart(Request $request)
     {	
+        $completed = ImportJob::whereProfileId(Auth::user()->profile->id)
+            ->whereService('instagram')
+            ->whereNotNull('completed_at')
+            ->exists();
+        if($completed == true) {
+            return redirect(route('settings'))->with(['errors' => ['You can only import from Instagram once.']]);
```
Codercise, line 30: Is the plan here that once the user has completed their Instagram import they should use Pixelfed forever?

trwnh: hmm. yeah, idk about this, it should be possible to import/export multiple times ideally. that would require duplicate checking and partial archives, but it's much better than saying that you can only import once.

dansup: @Haydos585 That is a good point, I will investigate a solution.

Codercise: You could try and limit it to once a month or something. I think with a service like Pixelfed if people are going to use it they will come, add some images, leave for 6 months, try it again, leave for 3 months, try it again, stick with it.

Having the same images imported multiple times is a problem too.

Perhaps something like `from_instagram` and `instagram_id` could be added to the Post table. Code can then figure out if something has been imported before.

dansup:
There are a few issues with the current implementation, so I will merge it for now without the routes so it can't be used until we can solve the issues @Haydos585 and @trwnh pointed out.

bjoernfan:
Hi. I set up my own Pixelfed instance yesterday and I'm very interested in using it on a regular basis. Say I, as a non-typical user in this scenario, am happy with a one-off import - would this existing code (+ routes) serve this need today?

aschrijver:
I'm a user in need of continuous updates. Our [Humane Tech Community](https://community.humanetech.com) has both PF and IG channels. I as fediverse fan wanna do more with PF [`@humanetech`](https://pixelfed.social/humanetech), while other members manage our [IG account](https://www.instagram.com/humane_tech_community/) (I'm not using that shit, but our members do their contribs there).

baarkerlounger:
Same question as bjoernfan - is there a way to use this for a one-off import on my hosted instance currently?

Rik44444:
Hi,
I wanted to ask, is this still on the agenda? Continuous sync with Insta would be amazing; single imports from insta would also already be a big pre. 
I found this, a closed Github item for this: https://github.com/pixelfed/pixelfed/issues/765 
and this item, which appears to be still active (https://github.com/pixelfed/pixelfed/pull/1046)
and this post, which appears to suggest it is ready to be launched: https://mastodon.social/@pixelfed/100087667113289786 , but i cannot find it in my instance (pixelfed.de)

PR #1105, opened 2019-04-02
Remove hover text-decoration for icon-only links

dermike:
This might just be personal taste, but here's a proposed change to remove text-decoration for links that are Font Awesome icons only (no text).

Mouse over on cog-icon before:
Screenshot 2019-04-02 at 10 06 09

Mouse over on cog-icon after:
Screenshot 2019-04-02 at 10 10 20


dansup:
Fixed in latest commits using ```text-decoration-none```. Thanks anyways!

PR #1124, opened 2019-04-05
Add COSTAR

dansup:
COSTAR (Confirm Object Sentiment Transform and Reduce) is a filter system for pixelfed inspired by the MRF (Message Rewrite Facility) feature in Pleroma. 

This PR adds basic COSTAR support with the following features:
- [x] Block instances 
Block instances at the AP level, preventing users from following users on blocked instances or from remote content from being stored

- [x] Auto CW instances
Apply a content warning to every post from this instance

- [x] Un-list instances from timelines
Remove posts from network timeline

- [x] Keyword filter
Block posts containing a specific keyword

- [x] Actor block
Block a specific remote actor/profile

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,3 +65,7 @@ HORIZON_DARKMODE=true
 #   php artisan optimize
 ACTIVITY_PUB=false
 REMOTE_FOLLOW=false
+
+CS_BLOCKED_DOMAINS='example.org,example.net,example.com'
+CS_CW_DOMAINS='example.org,example.net,example.com'
+CS_UNLISTED_DOMAINS='example.org,example.net,example.com'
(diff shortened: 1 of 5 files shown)


trwnh:
this needs to be designed in a granular way so that policies can be added or removed one-by-one:

scopes:
- [ ] user
- [ ] domain/site
- [ ] post (by hashtag/keyword or by metadata like `sensitive`, maybe?)

policies:
- [ ] apply auto-cw
- [ ] unlist from public timelines
- [ ] reject media
- [ ] reject statuses
- [ ] reject comments
- [ ] reject follows

dansup:
@trwnh Yes that is the long term goal, for now a few ```.env``` variables can perform basic filtering.

The way I see it, it would be easy to replace ```.env``` data source with more complex logic like storing rules in memory (redis) and in the database and then transforming them into the proper context for each function.

PR #1251, opened 2019-05-02
2) prepare .env files for instragram import

shleeable:


dansup:
Instagram import is not ready to be enabled

PR #1259, opened 2019-05-03
add doubletap to like

talon:
I attempted to add the `vue-touch@next` npm module. I don't know where it should be `Vue.use`'d but I put it in `bootstrap.js` for now. Assuming the plugin registers then there are also now `v-on:doubletap="likeStatus(status, $event)"` directives for all the status containers I could find.

I'm new to the project so I haven't been able to test this locally yet but perhaps someone can help me finish this PR

PR #1277, opened 2019-05-07
Refactor #1259, add double tap to like and deprecate lightbox

dansup:


talon:
THANK YOU! Now @dalerale will use pixelfed! 😝

PR #1313, opened 2019-05-21
Update docker-compose.yml - Adding restarts

shleeable:


docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
     # Comment to use dockerhub image
     build: .
     image: pixelfed
+    restart: always
```
techknowlogick: perhaps `restart: unless-stopped` instead? this way it'll stop if asked.

shleeable: I'm happy with that.

dansup (approved):

PR #1392, opened 2019-06-11
fix Redis notification exception

Findus23:
Hi,

Without this change the request to `api/v1/notifications` fails for me.

```json
{
    "message": "Non-static method Redis::zRangeByScore() cannot be called statically",
    "exception": "Symfony\\Component\\Debug\\Exception\\FatalThrowableError",
    "file": "/var/www/pixelfed/app/Services/NotificationService.php",
    "line": 24,
    "trace": [
    ]
}
```

I'm using PHP 7.3.4-2 with the php-redis module.

### DIFF
--- a/app/Services/NotificationService.php
+++ b/app/Services/NotificationService.php
@@ -2,7 +2,8 @@
 
 namespace App\Services;
 
-use Cache, Redis;
+use Cache;
+use Illuminate\Support\Facades\Redis;
 use App\{
 	Notification,
 	Profile


dansup:
@Findus23 Hello, please uninstall/disable the phpredis PECL extension as it is conflicting with the predis library pixelfed uses.

SharkyRawr:
> @Findus23 Hello, please uninstall/disable the phpredis PECL extension as it is conflicting with the predis library pixelfed uses.

Hi, sadly this is not an option for me as I run other software that requires it. Is there any other way to get pixelfed to work properly?

PR #1467, opened 2019-06-27
Update Privacy Settings

dansup:
Allow users to bulk manage followers during the process of making their account private.


jantheodor:
I would like to be able to alter the privacy settings of my images after there publication.

Issue #307, opened 2018-06-20
Idea: More tabs on profiles


trwnh:
Per IRC: we have two tabs on profiles currently:
- Posts: this shows all posts made by an account
- Saved: this shows posts you have bookmarked, and is only visible to you

Proposals for additional tabs:
- Collections: see #270
- Boosts: either make a view only for boosts/reshares, or alternatively change "Posts" to "{user}'s Posts" / "Posts and shares"
- Comments: view comments you've made on other people's posts. I would personally propose making this a private tab like Saved.
- A more general "updates" timeline for text-only posts, similar to iTunes Ping

```
[22:08:59]  <trwnh> dansup: Summary posts sounds interesting... as for pinned
                    posts though, I think it would be better to implement a
                    "highlight reel" option. People would highlight a post and
                    it would appear as a separate tab on their profile
[22:09:08]  <trwnh> Either that or have dynamic collections ;)
[22:09:24]  <trwnh> The abstraction from one to the other is probably not that
                    hard
[22:09:27] <dansup> trwnh: interesting!
[22:10:13]  <trwnh> dansup: The way I envision it right now is that profiles
                    need a third tab, and if that tab is Collections, then there
                    could be a Highlights collection
[22:10:55]  <trwnh> I think I might make a mockup of that later
[22:13:15] <dansup> trwnh: I think tabs are ok if we implement it right
[22:13:28] <dansup> collections should be called albums and show posts with more
                    than 1 media
[22:13:34]  <trwnh> dansup: it just feels unbalanced with only two tabs
[22:13:44]  <trwnh> especially if the profile is 3-column thumbnails
[22:13:56] <dansup> yeah, but think about IG
[22:14:12]  <trwnh> ehh
[22:14:25]  <trwnh> consideration to IG should be given but not copied 100%
[22:14:25]    <lcp> eh, class I made for it will adapt to any amount of tabs
[22:14:26] <dansup> if this were twitter/mastodon it would have 3 tabs. Posts, 
                    Posts w/ replies, Media
[22:14:49]  <trwnh> I don't know if it would be useful to show posts with
                    multiple media in its own tab
[22:14:50]    <lcp> you can make 20 tabs >D
[22:14:57]  <trwnh> That doesn't feel like it warrants its own tab
[22:15:27]  <trwnh> Arbitrary collections, though? A highlight reel? That would 
                    be fundamentally different than posting things
[22:15:34]    <lcp> it would probably be: posts, comments, boosts
[...]

PR #1511, opened 2019-07-17
Add Collections Feature

### LINKED ISSUE #270: Idea: Dynamic albums / Collections
While it may be possible to delete images selectively out of a multipost, it would probably be better to instead implement an album system where posts can be removed and added to separate collections or categories. For example, an album of all posts containing concert photography, or for a specific event, etc.

Potential considerations:
- Backend would treat these richer albums as an `OrderedCollection` of `Image` objects, perhaps `Note` attached
- At minimum, one level of nesting should be supported; otherwise, a separate category feature should be implemented as well. (More nesting is entirely arbitrary and would add complexity)
- The "Collections" / "Categories" feature(s) might live as extra tabs on the profile, next to "Posts" and "Saved".

dansup:
Adds Collections, the ability to curate many local/public posts in a single collection.

- [x] Add Models, Controllers, Migrations
- [x] Add collection api/web routes + add dynamic rate limit middleware
- [x] Add CollectionComponent.vue + collections.js
- [x] Add CollectionCompose.vue + collectioncompose.js
- [x] Add Collection rate limits
- [x] Update Profile.vue to display Collections
- [ ] ~~Update CollectionCompose.vue based on [feedback](https://mastodon.social/@pixelfed/102454003681798637)~~ **will ship in later version**

Fix #270 #307


jcarnat:
Thanks for the awesome work!

Do you already plan to implement the following features on Collections ?
- select a specific picture as the collection title backgroud.
- make it possible to change pics order in collections.
- make it possible to browse and share a whole collection from ActivityPub clients.

PR #1574, opened 2019-08-07
L10N: Fix German umlauts

es-ka:
Most German umlauts were HTML encoded in the source files, for example 'Ö' => `&Ouml;`
This doesn't work for me, as the characters were displayed with encoding. After removing the encoding the strings are displayed as expected. At least I checked the results for the strings in `auth.php` and `navmenu.php`.


monkeyless (approved):
Finde ich gut :)

dansup (approved):

PR #1650, opened 2019-08-30
v0.10.1

dansup:


app/Transformer/ActivityPub/StatusTransformer.php
```diff
@@ -54,6 +54,13 @@ public function transform(Status $status)
             ];
           }),
           'tag' => [],
+          'location' => $status->place_id ? [
+              'type' => 'Place',
+              'name' => $status->place->name,
+              'longitude' => $status->place->long,
+              'lattitude' => $status->place->lat,
```
jplock: Should these be “latitude”?

dansup: Fixed in a30575e

PR #1704, opened 2019-09-13
Default word-break styling on post comments

trevdor:
To my eye, it appears `word-break: break-all` gives jarring word breaks at least for Latin alphabet languages. Removing it seems to smooth things over. 
If the existing pseudo-justified text feel is desirable, I could add `text-align: justify;` (see last screenshot).

#### Current
break-all

#### Proposed change, as written
no word-break property

#### Proposed change + justified text
Screen Shot 2019-09-13 at 10 48 59 AM


trwnh (approved):
Looks good to me; I would say it's fine as-is IMO

dansup (approved):

Issue #1737, opened 2019-10-01
Allow Translation For Places


okpierre:
At the moment 'Places' is just direct link. 

Allowing translation will give instances their own language to match with rest of site

ahstro:
This can be closed now, since fix is merged :)

PR #1739, opened 2019-10-01
Update site.php

okpierre:
@dansup this should help with translation https://github.com/pixelfed/pixelfed/issues/1737

### DIFF
--- a/resources/lang/en/site.php
+++ b/resources/lang/en/site.php
@@ -14,5 +14,6 @@ return [
 	'selectLocale'  => 'Select one of the supported languages',
 	'contact'		=> 'Contact',
 	'contact-us'	=> 'Contact Us',
+    'places'        => 'Places',
 
 ];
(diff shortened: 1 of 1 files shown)


resources/lang/en/site.php
```diff
@@ -14,5 +14,6 @@
 	'selectLocale'  => 'Select one of the supported languages',
 	'contact'		=> 'Contact',
 	'contact-us'	=> 'Contact Us',
+    'places'        => 'Places',
```
ahstro: Looks like the whitespace is off here. Not sure if there's a space or two too many or if there is an issue with tabs vs. spaces, but you probably want to fix this :)

ahstro (commented):
I don't have write access, so don't count this as a proper review, but I'm assuming it'd be better if this was combined with #1738 and the whitespace issue is fixed, but otherwise it looks good to me :)

resources/lang/en/site.php
okpierre: Arrgh yeah tabs/white space always does it. Will clean it up. Thanks!

dansup (approved):

PR #1805, opened 2019-11-01
improve German translation

Strubbl:


Strubbl:
small German translation update
//CC @timmyRS @nin-tan @t-aus-m @es-ka @ahstro

dansup (approved):

PR #1809, opened 2019-11-03
Remove unecessary docker stuff

Strubbl:
@Shleeble i removed the line you questioned and also the providing .env to the redis container. ok?


dansup:
We really need to separate the docker stuff from the main repo.

dansup (approved):

Strubbl:
@dansup yes, you are right. Just kick-off https://github.com/pixelfed/docker and we can move that stuff.

PR #1941, opened 2020-01-17
Don't always explode the TRUST_PROXIES variable

asonix:
Since the trustedproxy middleware checks the proxy variable against '*' and '**' directly, pixelfed should preserve those values as strings, rather than exploding them into arrays of strings.

This solves my issue with a load-balancer in front of pixelfed.


config/trustedproxy.php
```diff
@@ -23,5 +31,5 @@
      * how many proxies that client's request has
      * subsequently passed through.
      */
-    'proxies' => explode(',', env('TRUST_PROXIES', '')),
+    'proxies' => $proxies,
```
dansup: Please inline the logic with something like ```'proxies' => strpos(env('TRUST_PROXIES'), ',') ? explode(',', env('TRUST_PROXIES')) : env('TRUST_PROXIES')```

dansup (changes requested):

dansup (approved):

config/trustedproxy.php
asonix: as discussed on matrix/irc, the underlying library supports comma-separated strings now. We can pass a string in directly.

PR #2015, opened 2020-02-15
BREAKING: Docker revision - Migrate to Postgres 12/Clean up dockerfile with comments

shleeable:
# BREAKING CHANGES:

1. This migrates from MySQL to Postgres 12 as the default DB. 
2. Move default image tool from `gd` to `imagick`.

--
Apache has almost been tested. I need to test and confirm the upgrade flow
We need to figure out a stable config for nginx.

fpm has never worked for me correctly as it currently is designed. I'm going to rebuild it after this is merged.

### Installation steps:
Confirm the .env.docker is setup correctly

> docker-compose up -d postgres redis
docker-compose run --rm app php artisan key:generate
docker-compose run --rm app php artisan passport:keys
docker-compose run --rm app php artisan migrate
docker-compose run --rm app php artisan horizon:install
docker-compose run --rm app php artisan import:cities
docker-compose up -d

docker-compose build --pull
docker-compose down app worker
docker-compose run --rm app php artisan migrate
docker-compose up -d


shleeable:
Note: https://github.com/dimitri/pgloader should be used to migrate from one to the other...
We'd just need to create a guide on migrating because it's going to be a little tricky

config/image.php
```diff
@@ -15,6 +15,6 @@
     |
     */
 
-    'driver' => env('IMAGE_DRIVER', 'gd'),
+    'driver' => env('IMAGE_DRIVER', 'imagick'),
```
dansup: This is a breaking change, I can't merge this PR without breaking instances that rely on GD

dansup (changes requested):

config/image.php
shleeable: ooooh I understand. that change is pointless as the env is in the `.env.docker`

dansup (dismissed):

PR #2027, opened 2020-02-16
Update Installer.php

shleeable:


app/Console/Commands/Installer.php
```diff
@@ -221,7 +220,7 @@ protected function updateEnvFile($key, $value)
         $envPath = app()->environmentFilePath();
         $payload = file_get_contents($envPath);
 
-        if ($existing = $this->existingEnv($key, $payload)) {
+        if ($existing == $this->existingEnv($key, $payload)) {
```
dansup, line 223: A single ```=``` was intentional as it sets the ```$existing``` variable used in the next line if ```existingEnv()``` is true.

PR #2029, opened 2020-02-16
Update PrettyNumber.php

shleeable:
for was empty.


app/Util/Lexer/PrettyNumber.php
```diff
@@ -31,9 +31,8 @@ public static function size($expression, $kb = false)
           ['B', 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'] :
           ['B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
         for ($i = 0; ($size / 1024) > 0.9; $i++, $size /= 1024) {
-        }
         $res = round($size, $precision).$units[$i];
-
+        }
```
dansup, line 35: This would break functionality, let me give you an example.

With a value of ```10000000```

### Before:
```bash
>>> PrettyNumber::size(10000000)
=> "10M"
```

### After: 
```bash
>>> PrettyNumber::size(10000000)
=> "9766k"
```

PR #2037, opened 2020-02-19
Use .env config for PF_USER_INVITES

Simounet:
Don't get why this was commented in 577dac98022f8e7107df64f570fd6c1072f32daa.
Have a nice day.


dansup:
@Simounet It's disabled because the feature is not ready yet.

Issue #1993, opened 2020-02-07
Code styles are inconsistent


dariusj18:
The projects has an `.editorconfig` file, however there are still commits like this  7355e94723428523aca4e47820c351ca6c246156 where the code changes are confusing because the spaces were changed to tabs.

Since this is a Laravel project I would advise following [PSR-12](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-12-extended-coding-style-guide.md).

[CircleCI](https://github.com/marketplace/circleci) or [GitHub actions](https://github.com/features/actions) could be used to automate testing.

Here's an example I found with a quick search of a project implementing Github actions for php code sniffing.  https://github.com/typo3-trainer-network/books/pull/38/files

shleeable:
It's easy to add a github action to check against these standards.

I've just done a scan and the PR to change up to PSR-1, then PSR-12 is going to effect every file ;)
So @dansup really needs to be keen to run these code standard changes, because his code standard is obviously his preference.

dansup:
Someone else contributed the ```.editorconfig``` file, we did use StyleCI at one point but it wasn't configured properly so I removed it.

I plan to tackle this in v0.12 (the big refactor), will keep this open until I fix!

svenluijten:
@dansup what's the status of this? As the project has gotten more contributors, I think it's smart to tackle this _now_ before it gets _too_ unwieldy and it can't be changed at all (if it hasn't yet). Enabling StyleCI should be pretty minor and only affect _new code_ that's contributed to the project, so you won't have a massive PR like #2064.

jippi:
https://github.com/pixelfed/pixelfed/pull/4921 +https://github.com/pixelfed/pixelfed/pull/4922 was intended to fix this :)

PR #2046, opened 2020-02-22
PSR2

shleeable:
starting #1993, this is a simple start to polish the code as per https://www.php-fig.org/psr/psr-2/ standard.

I've got another batch of PRs but this will help us prepare - Plenty more where this came from.


shleeable:
This is obviously a massive change... but they're all pretty boring

- single line imports over array of common namespaces.
- Fixes whitespace.
- else if -> elseif

dansup:
As discussed on IRC, this PR is too big. We'll be using [StyleCI](https://styleci.io/) to perform these changes in the near future!

PR #2157, opened 2020-04-29
Update Inbox

dansup:


app/Util/ActivityPub/Inbox.php
```diff
@@ -293,8 +293,6 @@ public function handleDeleteActivity()
         if(!isset(
             $this->payload['actor'], 
             $this->payload['object'],
```
Strubbl, line 295: is the comma here misplaced, which makes federation stop working? see also my comments in pixelfed-dev chat.

PR #2211, opened 2020-05-26
Add Instagram Import

dansup:
## Adds Instagram Import 🎉 🥳

Edit .env and add ```IMPORT_INSTAGRAM=true``` to enable.

Default limit is 100 posts, to change edit ```IMPORT_INSTAGRAM_POST_LIMIT``` .env variable with an integer.

Users can only perform one import during the beta, admins will be able to clear previous imports to allow them to try again if any issues arrise.


vinyll:
Hey @dansup,

How can we access that feature on https://pixelfed.social/? Is it enabled?

PR #2236, opened 2020-06-08
fixes cors package moved to new org

der-On:


dansup:
Added in 344b0c9 + #2220

PR #2318, opened 2020-07-10
Change response of bookmarks

### LINKED ISSUE #2317: Bookmarks API broken JSON output
Doing a call on api/v1/bookmarks returns invalid JSON which can't be parsed.
Compare with api/v1/favourites for instance which is totally fine.

Experienced on pixelfed.social

Screenshot from a response from a rest client - first one is fine (favourites), second one is broken (bookmarks).

favourites-response

bookmarks-response

swentel:
This fixes https://github.com/pixelfed/pixelfed/issues/2317 - screenshots are on that issue

I've basically copy/pasted the way favourites method gets the data and returns the response.

Note: haven't tested with actual data, don't have a running instance yet on my local machine, so make sure to double check :)

### DIFF
--- a/app/Http/Controllers/Api/ApiV1Controller.php
+++ b/app/Http/Controllers/Api/ApiV1Controller.php
@@ -1876,9 +1876,9 @@ class ApiV1Controller extends Controller
     /**
      * GET /api/v1/bookmarks
      *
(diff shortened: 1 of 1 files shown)


dansup:
Fixed in #2324

PR #2354, opened 2020-07-27
Add MediaBlocklist feature

dansup:


ThisIsMissEm:
I wouldn't recommend using cryptographic hashes to implement this feature, but instead using Perceptual Hashes – cryptographic hashes are easily defeated by changing the image slightly (all it'd take is 1px of the image to change for it's hash to change)

Perceptual Hashes are what we use in systems designed to detect CSAM, revenge porn, and violent gore content, as they are a hash of what's visible/interesting in the image.

What Facebook uses is called TMK+PDQF, a reference implementation is here: https://github.com/facebook/ThreatExchange/tree/main/tmk

The one used by Thorn, NCMEC, etc, for CSAM is proprietary & closed source (as it's an arms race between abusers & criminals and the perceptual hashing algorithms)

You could probably get pretty far by using pHash or similar, and it'd be a lot more resilient compared to cryptographic hash matching, as it'd catch resized images, images with colour changes, etc.

PR #2377, opened 2020-08-12
Use json_decode in bookmarks api response

swentel:
Because now, the json is double encoded, see https://github.com/pixelfed/pixelfed/issues/2317 for more background


dansup:
Fixed in #2494

Issue #2202, opened 2020-05-24
Rate limit requests to other ActivityPub implementations


chendo:
Our Mastodon instance is seeing occasional spikes of 500 requests in a second from "Pixelfed/0.10.9; +https://pixelfed.social" for the same URL for a user on our instance that has been deleted. The user has been deleted so it returns 410 Gone.

Pixelfed should consider rate limiting requests to other ActivityPub instances, as well as consider client-side caching of these responses. I can supply log files of this issue privately if you'd like.

chendo:
Observed another 653 requests for the exact same user over the space of 3 seconds. We've mitigated this by caching 410 responses but our DoS mitigation has caught pixelfed.social multiple times.

dansup:
Fixed!

PR #2411, opened 2020-09-22
Cache the results of profileFetch for 1 minute

chendo:
This should address #2202 where account suspensions on other ActivityPub implementations (Mastodon in this case) causes Pixelfed to attempt to fetch the user per Status that was deleted due to account suspensions.

I don't have a PHP environment so I can't test this myself unfortunately. I assume `Cache::remember` cannot store nulls.


dansup:
Shipped in #2454

PR #2511, opened 2020-12-15
remove dupl. lines avoid error on avatar-update

xundeenergie:
duplicate lines and trailing || removed -> no error on update avatar


cyberkeiler:
Can confirm that. This fixes the issue.

PR #2513, opened 2020-12-17
Fix usertag link

### LINKED ISSUE #2512: Link to federated profile on comments lead to wrong subpath
When i add a comment with a mentioned user (for example from my own mastodon-instance), the username gets blue and is linked to https://japix.schuerz.at/@jakob@mastodon.schuerz.at?src=rph
grafik
grafik

That's correct.
Reloading the page changes the link. Now it points to https://japix.schuerz.at/discover/tags/@jakob@mastodon.schuerz.at?src=rph which is wrong and leads to 
grafik

xundeenergie:
fixes https://github.com/pixelfed/pixelfed/issues/2512

### DIFF
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -121,10 +121,7 @@ class AvatarController extends Controller
         $avatar = $profile->avatar;
 
         if( $avatar->media_path == 'public/avatars/default.png' || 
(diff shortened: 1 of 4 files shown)


xundeenergie:
What's wrong with the solution for the wrong discover/tags issue???

trwnh:
@dansup? lgtm but i'm not familiar with the code

PR #2536, opened 2020-12-28
fix wrong link in usertag

xundeenergie:
When mention a user (foreign user, for example a mastodon-user with @username@mastodon.instance) the link in comment points to the internal destination /discover/tags/username and should point to /@username@mastodon.instance.

This PR fixes the Path in Comment.


xundeenergie:
wrong blanks (space instead of tab)

PR #2537, opened 2020-12-28
allow user create with flags in CLI

### LINKED ISSUE #2567: Can't automatically create a user/admin when spinning up
Would be nice to be able to automatically have an admin account, ready to use when spinning up a pixelfed instance..

rationale:
For some users it may be easier/safer to moderate the instance from the UI rather than the command line, so less stuffing around in the command line would be better.. right?

fix: 
1. to have an admin account already spun up or..
2. to be able to set the admin account (using the command line) automatically during spin up, so the command line doesn't need to be used (for that particular chore)

ghost:
closes #2567 
Allows the server owner to set/create a user, without needing to go through the interactive form.

This is done by using flags, eg. php artisan user:create --username=voryzen

possible flags are:
{--name=} {--username=} {--email=} {--password} {--is_admin=}

If a flag is not used, the interactive form will ask.

Usecase:
setting an accounts automatically, when using docker-compose or another automatic deploying service, especially the all important admin account!

### DIFF
--- a/app/Console/Commands/UserCreate.php
+++ b/app/Console/Commands/UserCreate.php
@@ -12,7 +12,7 @@ class UserCreate extends Command
      *
      * @var string
      */
-    protected $signature = 'user:create';
+    protected $signature = 'user:create {--name=} {--username=} {--email=} {--password} {--is_admin=}';
 
     /**
      * The console command description.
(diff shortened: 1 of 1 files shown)


ghost:
what do you reckon, @dansup ?

ghost:
no feedback? @dansup

PR #2538, opened 2020-12-28
Fix link in user-tag in comment

xundeenergie:
Usertag in comment should point to /@username@instance, but it points to /discover/tags/@username@instance.

This PR fixes the link in Comments.


ghost:
@xundeenergie 
Hey mate,

Just so you know.. you are supposed to push to staging, instead of dev.

HTH

dansup:
Fixed in #2571

PR #2580, opened 2021-01-24
Remove auto-generated terms of service

### LINKED ISSUE #2079: Remove Default TOS if Out of Scope
It seems that any issues regarding the currently included TOS have been closed as outside of the scope of the Pixelfed project. (https://github.com/pixelfed/pixelfed/issues/1060#issuecomment-577543261) However, it also seems irresponsible to provide a default TOS (which many users will likely trust to be safe to use) that contains so many problematic statements.

One particular section I want to call attention to is the following:
>[U]nder this license you may not: . . . transfer the materials to another person or "mirror" the materials on any other server.

This seems to prohibit federation with any Pixelfed server by default.

While the likelihood of a Pixelfed instance admin ever pursuing any legal action regarding this statement (or any others in the default TOS) is debatable, it should probably be removed to completely avoid the legal risks involved.

My point is not that any individual part of the default TOS is an issue. Rather, the entire default TOS is unsuitable for ANY Pixelfed instance, and should therefore be removed.

obligatory IANAL

trwnh:
Having incorrect TOS is worse than having no TOS. Do also note that e.g. Mastodon's terms are simply its privacy policy, and there is already a privacy.blade.php template.

Fix #2079

### DIFF
--- a/resources/views/site/terms.blade.php
+++ b/resources/views/site/terms.blade.php
@@ -10,38 +10,6 @@
 				{!! $page->content !!}
 				@else
 				<div class="terms">
-					<h5 class="font-weight-bold">1. Terms</h5>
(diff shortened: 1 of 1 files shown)


xzed (approved):

dansup:
Admins can edit TOS, add/edit rules and set a custom description in the latest commits!

PR #2581, opened 2021-01-24
feat: add flags to user create cli command

ghost:
Allows the server owner to set/create a user, without needing to go through the interactive form.

This is done by using flags, eg. php artisan user:create --username=voryzen

possible flags are:
{--name=} {--username=} {--email=} {--password} {--is_admin=}

If a flag is not used, the interactive form will prompt the user.

Usecase:
setting an accounts automatically, when using docker-compose or another automatic deploying service, especially the all important admin account!


dansup:
Fixed in #2571 with more options

ghost:
Nice work, @dansup 

Lovin' it

PR #2592, opened 2021-01-29
Update StatusTransformer, add blurhash

dansup:


trwnh:
shouldn't blurhash be computed at the API level instead of the federation level?

PR #2607, opened 2021-02-01
Update StatusLabelService.php

### LINKED ISSUE #2606: Labels not applied correctly for covid19
incorrect label when post includes beer
post does not apply to pandemic tag

Pixelfed v0.10.10

okpierre:
Fixes https://github.com/pixelfed/pixelfed/issues/2606

removed corona to avoid adding WHO link to posts that contains beer
removed typo for pandemic
added covid19


dansup:
Would rather have false positives than false negatives, at least for now. And `plandemic` is not a typo, we should add `pandemic` as well.

PR #2642, opened 2021-02-09
Create directory in InstagramImport

### LINKED ISSUE #2612: v0.10.10, ErrorException when importing from Instagram
Running 0.10.10 version.

When importing photos from Instagram, downloaded to local directory and selected from there, Horizon throws this error:

ErrorException: rename(/var/www/pix/pixcat/storage/app/import/9c046093-3b1c-4671-856f-eb9f13c3bbe7/murfcouOtLKQdiM3Typh6wdbB4MkwJwizGYls9Bb.jpg,/var/www/pix/pixcat/storage/app/public/m/01425e939359e081ec015715d0dd00fc7d06754e/1d580e6fb305dc377450e8f0146c6d7dbcfa5dc9/5c2e9b165caa32e7cce3260906d0026e.jpg): No such file or directory in /var/www/pix/pixcat/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php:290

Target directory does not exists so seems a create dir permissions problem. I followed exactly the instructions [to set them right](https://docs.pixelfed.org/running-pixelfed/installation.html#set-correct-permissions) but ErrorException keeps throwing above error.

Even tried a new fresh install in a new directory but nothing changed.

stelzch:
Should fix #2612

### DIFF
--- a/app/Jobs/ImportPipeline/ImportInstagram.php
+++ b/app/Jobs/ImportPipeline/ImportInstagram.php
@@ -101,7 +101,11 @@ class ImportInstagram implements ShouldQueue
 
                 $path = storage_path("app/$importData->path");
                 $storagePath = "public/m/{$monthHash}/{$userHash}";
-                $newPath = "app/$storagePath/$filename";
+                $dir = "app/$storagePath";
+                if(!is_dir(storage_path($dir))) {
+                    mkdir(storage_path($dir), 0777, true);
+                }
+                $newPath = "$dir/$filename";
                 $fs->move($path,storage_path($newPath));
                 $path = $newPath;
                 $hash = \hash_file('sha256', storage_path($path));


app/Jobs/ImportPipeline/ImportInstagram.php
```diff
@@ -101,7 +101,11 @@ public function handle()
 
                 $path = storage_path("app/$importData->path");
                 $storagePath = "public/m/{$monthHash}/{$userHash}";
-                $newPath = "app/$storagePath/$filename";
+                $dir = "app/$storagePath";
+                if(!is_dir(storage_path($dir))) {
+                    mkdir(storage_path($dir), 0777, true);
```
dansup, line 106: Please remove the second and third parameters of the `mkdir` function

dansup (changes requested):

app/Jobs/ImportPipeline/ImportInstagram.php
stelzch: I think the third parameter is necessary in order to correctly create the parent directory.
Will check however.

stelzch: I tried removing the second and parameter, but the import did not work afterwards. The second parameter could be omitted, PHP 8 named arguments make it possible:

```php
mkdir(storage_path($dir), recursive: true);
```

However, since PHP versions <8.0 are also targeted, I would leave it just as is.

@dansup

aftek, line 106: isn't 0777 a bit overkill though ?
```suggestion
                    mkdir(storage_path($dir), umask(), true);
```

just a thought,
new comer here looking around :)
I'd make sure it's not more than 0750 if hosting myself.

stelzch: The way I read the [documentation of mkdir](https://www.php.net/manual/en/function.mkdir.php), the mode parameter is already modified by the current umask. Also, 0777 is the default argument if no other is supplied. I quickly tested it from the command line, and with umask set to 022, `mkdir("dir", 0777)` creates a directory with mode 0755.

But sure, 0750 would work, too. Maybe the maintainers could state their preference.

dansup (approved):

PR #2644, opened 2021-02-10
Clean up artisan commands

trwnh:
Make some things more consistent and easier to find, better grouping, etc

Changes included:

- Rename story:gc to gc:story because there are already other gc: tasks
- Rename media:gc to gc:media
- Rename media:optimize to fix:optimize
- Rename media:fix to fix:media
- Rename regenerate:thumbnails to fix:thumbnails
- Rename video:thumbnail to fix:thumbvideo
- Rename status:dedup to fix:dedup


dansup:
Renaming commands is too risky, it could break things with little benefit. Going to close this for now.

PR #2653, opened 2021-02-13
Enable multiple file select when creating a new post

Johnny-Malizia:
This should allow multiple files to be selected for upload and the uploading process seems to already account for this.


dansup:
We used to do this but it was confusing to some users.

PR #2719, opened 2021-04-07
Add Network Timeline

dansup:
Adds support for Network Timeline.

To enable this you must add `PF_NETWORK_TIMELINE=true` to `.env` file and run `php artisan config:cache`


t3rr0rz0n3:
Please, add this parameter in documentation! https://docs.pixelfed.org/technical-documentation/env.html

PR #2868, opened 2021-07-27
Fix PWA implementation to support Gecko-based browsers, using html link tag

nogafam-es:
In order to implement PWA successfully (so Browsers have the option to use Pixelfed full screen as an App), you need to add this line into the main / (root) view.

Firefox (for example) does not achieve to use PWA currently. I guess this need is only for Firefox (Gecko-based browsers).

Source: https://developer.mozilla.org/en-US/docs/Web/Manifest


nogafam-es:
This is a video i made (in Spanish), that shows the issue https://pixel.nogafam.es/p/bofh/325755983957725184

dansup (approved):

PR #2934, opened 2021-09-15
fix 'powered by pixelfed' link to be HTTPS

vszakats:


vszakats:
This insecure link pointing to `pixelfed.org` now appears in a prominent place in _every_ timeline. Some browsers have the option to disable HTTP pages. In those browsers this link will not work, in the rest of the cases an insecure (MitM-able) request will be made, which also isn't ideal given that everything else is correctly done via HTTPS.

Is there some issues, concerns or things to do regarding this PR?

vszakats:
@dansup: Any love for (or comment on) this one?

dansup (approved):

PR #3015, opened 2021-11-21
Primary key installation issue fix & some SQL queries were returning null

samihsoylu:
This merge request fixes the following issues:
* DigitalOcean managed MySQL 8.0 databases throw an exception during `php artisan migrate --force`. It is a configuration requirement that each database table must have a primary key. Since it appears some tables do not have a primary key, it makes sense to disable the `sql_require_primary_key` setting.
* Some pages do not work out of the box. An example can be seen in [issue 137](https://github.com/pixelfed/support/issues/137). I found out that this behavior was caused by SQL queries returning `null` when a provided record with an `id` was not found in the database. To address this, the provided `$id` parameter needed to be typecasted to an integer. When testing, the SQL query returned an actual result.
* The admin profiles section where you have multiple options such as enforcing content warning, unlisting posts, etc. did not function out of the box. I added hidden fields to the appropriate template to provide an `action` when the post request is made.


samihsoylu:
Closed due to no response from the repository owner

Issue #2465, opened 2020-11-29
Artisan and user functions doesn't work (QueryException)


mauri5hut:
I cannot use Artisan and the user functions. First was trying to make user admin, but I got the same error when using the "show" function:

-------------------------------------------------------------------------------
 Illuminate\Database\QueryException 

SQLSTATE[22P02]: Invalid text representation: 7 FEHLER:  ungültige Eingabesyntax für ganze Zahl: »myuser« (SQL: select * from "users" where ("username" = myuser or "id" = myuser) and "users"."deleted_at" is null limit 1)

at vendor/laravel/framework/src/Illuminate/Database/Connection.php:671
667|         // If an exception occurs when attempting to run a query, we'll format the error
668|         // message to include the bindings with SQL, which will make this exception a
669|         // lot more helpful to the developer instead of just the database's errors.
670|         catch (Exception $e) {
**671|             throw new QueryException(**
672|                 $query, $this->prepareBindings($bindings), $e
673|             );
674|         }
675| 

+14 vendor frames 
15  app/Console/Commands/UserAdmin.php:42
Illuminate\Database\Eloquent\Builder::first()

+14 vendor frames 
30  artisan:37
Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))

-------------------------------------------------------------------------------

PF Version: 0.10.9
Using Postgres as Database.

cydactyl:
Having this same issue, seems to be trying to compare an int with a string, but not familiar enough with php to fix.

ghost:
Having basically the same issue... I want to promote a user to admin,but artisan is failing. In fact the SQL statement is messed up: 

Artisan tries to do this: 

`select * from "users" where ("username" = ij or "id" = ij) and "users"."deleted_at" is null limit 1` 

First, there are some syntax issues: 

1. In PostgreSQL column names are not within "" quotes. In fact, I don't think that Postgres likes double quotes at all. 
2. instead of the columns the values/strings should be quoted like username='ij'

When looking at the table definition, you can also spot another issue: 

```pixelfed=# \d users
                                             Table "public.users"
      Column       |              Type              | Collation | Nullable |              Default
-------------------+--------------------------------+-----------+----------+-----------------------------------
 id                | bigint                         |           | not null | nextval('users_id_seq'::regclass)
 name              | character varying(191)         |           |          |
[...]

PR #3267, opened 2022-02-19
Change pixel/app/Console/Commands/UserAdmin.php to fix a bug

Gxnfvhg:
Replacing "id" with "name" fixes a bug mentioned in #2465 

`<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use App\User;

class UserAdmin extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'user:admin {name}';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Make a user an admin, or remove admin privileges.';

    /**
     * Create a new command instance.
     *
     * @return void
     */
    public function __construct()
    {
        parent::__construct();
    }

    /**
     * Execute the console command.
     *
     * @return mixed
     */
    public function handle()
    {
        $name = $this->argument('name');
        $user = User::whereUsername($name)->orWhere('name', $name)->first();
        if(!$user) {
            $this->error('Could not find any user with that username or id.');
            exit;
        }
        $this->info('Found username: ' . $user->username);
        $state = $user->is_admin ? 'Remove admin privileges from this user?' : 'Add admin privileges to this user?';
        $confirmed = $this->confirm($state);
        if(!$confirmed) {
            exit;
        }

        $user->is_admin = !$user->is_admin;
        $user->save();
        $this->info('Successfully changed permissions!');`


Gxnfvhg:
Damn, i really messed up there with Merging 72 other pull requests, closing this ahah !

PR #3293, opened 2022-03-06
Fix error when collection has no posts

### LINKED ISSUE #3289: Call to a member function thumb() on null
Pixelfed v0.11.2

`Call to a member function thumb() on null {"userId":811,"exception":"[object] (Error(code: 0): Call to a member function thumb() on null at /pixelfed/app/Http/Controllers/CollectionController.php:213)`

stacktrace https://pastebin.com/JwPQbZXY

rm-yakovenko:
Should fix #3289


dansup:
Fixed in #3295

Issue #3049, opened 2021-12-18
Specifies "application/x-www-form-urlencoded" in the Content-Type Header


toddsundsted:
It looks like Pixelfed is sending ActivityPub activities with a content type of "application/x-www-form-urlencoded" instead of 'application/ld+json; profile="https://www.w3.org/ns/activitystreams"' or even "application/activity+json". Unless explicitly specified, PHP cURL uses either "multipart/form-data" or "application/x-www-form-urlencoded", depending on the data.

The correct header could be added in:
https://github.com/pixelfed/pixelfed/blob/dev/app/Services/ActivityPubDeliveryService.php#L44

or in the signature generation code:
https://github.com/pixelfed/pixelfed/blob/dev/app/Util/ActivityPub/HttpSignature.php#L17

I could wrong, but I think the relevant documentation is here: https://www.php.net/manual/en/function.curl-setopt.php#refsect1-function.curl-setopt-notes.

spctrlth:
Bump. This seems to affect federation with Pleroma, see https://git.pleroma.social/pleroma/pleroma/-/issues/2790#note_90849

dansup:
Fixed in #3411

PR #3410, opened 2022-05-02
Fix request Content-Type for ActivityPubDeliveryService

noellabo:
Fix https://github.com/pixelfed/pixelfed/issues/3049


trwnh:
would it not be more correct to use `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` instead of `application/activity+json`?

noellabo:
This is the same header that Mastodon adds in Delivery, but there may be something more appropriate.

BTW, I've checked the code more broadly and found similar issues in other places.
I'll close this pull request once and suggest again.

dansup:
Fixed in #3411

PR #3411, opened 2022-05-02
Add Content-Type and User-Agent for activity delivery

noellabo:
Fix https://github.com/pixelfed/pixelfed/issues/3049

Add the appropriate Content-Type and User-Agent to the header when delivering the Activity. 

In addition to Pleroma, it solves the problem of delivery not reaching servers that are blocking inappropriate requests by web application firewalls. (An example of WAF is mstdn.jp)


trwnh:
looks fine to me in theory, but i do not know enough php to verify 100% so i am requesting a review from dansup

dansup (approved):

dansup:
@noellabo Thank you!

hikerus:
Do you implement this fix in your production server pixelfed.social very soon?

PR #3436, opened 2022-05-11
SSO OIDC client support

osresearch:
This is a *very early* draft set of changes to integrate Single Sign On using OpenID Connect into PixelFed.  This hacks together the [GCS OIDC Client](https://bitbucket.org/cabinetoffice/oidc-client/src/master/) library with some glue to create user accounts based on the values received from the SSO server (Keycloak in my tests).

Todo:
* [x] `isAdmin` should be set based on the roles in the SSO account
* [x] Full name, email and username (and admin) need to be updated on each login
* [x] UI needs to hide normal login if OIDC is enabled
* [ ] UI needs to be themed so that it isn't just a bare `<a href>` link
* [x] Changes squashed
* [ ] Documentation updated

Caveats: I don't know anything about PHP, SSO, Artisan, Composer etc so please feel free to make things more idiomatic.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -148,3 +148,9 @@ TRUST_PROXIES="*"
 ## Passport
 #PASSPORT_PRIVATE_KEY=
 #PASSPORT_PUBLIC_KEY=
+
+# Login with OpenID Connect for Single-Sign-On registration and authentication
+#OIDC_ENABLED=true
+#OIDC_CLIENT_ID=pixelfed
+#OIDC_CLIENT_SECRET=RANDOM-SECRET-HERE
+#OIDC_PROVIDER_URL=https://login.real.domain/realms/realm-name
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,8 +22,7 @@ services:
     env_file:
       - .env.docker
     volumes:
-      - app-storage:/var/www/storage
-      - app-bootstrap:/var/www/bootstrap
+      - ./data/app-storage:/var/www/storage
       - "./.env.docker:/var/www/.env"
     networks:
       - external
@@ -43,8 +42,7 @@ services:
     env_file:
       - .env.docker
     volumes:
-      - app-storage:/var/www/storage
-      - app-bootstrap:/var/www/bootstrap
+      - ./data/app-storage:/var/www/storage
     networks:
       - external
       - internal
@@ -63,7 +61,7 @@ services:
     env_file:
       - .env.docker
     volumes:
-      - "db-data:/var/lib/mysql"
+      - ./data/db-data:/var/lib/mysql
 
   redis:
     image: redis:5-alpine
@@ -71,15 +69,9 @@ services:
     env_file:
       - .env.docker
     volumes:
-      - "redis-data:/data"
+      - ./data/redis-data:/data
     networks:
       - internal
-
-volumes:
-  db-data:
-  redis-data:
-  app-storage:
-  app-bootstrap:
 
 networks:
   internal:
--- a/oidc-client/src/Auth/OIDCUserProvider.php
+++ b/oidc-client/src/Auth/OIDCUserProvider.php
@@ -1,0 +1,130 @@
+<?php
+
+
+namespace GCS\OIDCClient\Auth;
+
+
+use Illuminate\Contracts\Auth\Authenticatable;
+use Illuminate\Contracts\Auth\UserProvider;
+
+
+class OIDCUserProvider implements UserProvider
+{
+    private function is_admin($user_info)
+    {
+	    if (!property_exists($user_info, "resource_access"))
+		    return false;
+	    if (!property_exists($user_info->resource_access, "pixelfed"))
+		    return false;
+	    if (!property_exists($user_info->resource_access->pixelfed, "roles"))
+		    return false;
+	    return in_array("admin", $user_info->resource_access->pixelfed->roles);
+    }
+
+    public function retrieveByInfo($user_info)
+    {
+        \Log::info(print_r($user_info, true));
+        $model = config('auth.providers.users.model');
+	$user = $model::where('email', $user_info->email)->first();
+	if ($user) {
+		// user exists, make sure the other fields match
+		$need_save = false;
+		if ($user->name != $user_info->name)
+		{
+			$user->name = $user_info->name;
+			$need_save = true;
+		}
+		if ($user->username != $user_info->preferred_username)
+		{
+			$user->username = $user_info->preferred_username;
+			$need_save = true;
+		}
+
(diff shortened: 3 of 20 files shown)


osresearch:
Getting the `admin` role from the `userinfo` works, although sudo mode attempts to ask for a password.  It looks like we should be able to set the `$request->session()->put('sudoTrustDevice', 1);` in the OIDC code, ~although I'm not sure how to get the session from that context.~ the guard class can set it.

ser:
Any chance to restart work on this? It is a very needed feature...

osresearch:
If someone who knows PHP can help out with the remaining tasks to better integrate it into the UI and login flow, it would be very appreciated!  I'll try to rebase it so that it at least usable with the current head.

One new todo item that has come up since I wrote the initial draft: support for the official app. I'm not at all familiar with that code tree, so I'm not sure what will be involved in supporting OIDC / OAuth2 with the mobile version.

.env.docker
```diff
@@ -145,4 +146,13 @@ TRUST_PROXIES="*"
 
 ## Passport
 #PASSPORT_PRIVATE_KEY=
-#PASSPORT_PUBLIC_KEY=
\ No newline at end of file
+#PASSPORT_PUBLIC_KEY=
+
+# OIDC for logins
+OIDC_CLIENT_ID=pixelfed
+OIDC_CLIENT_SECRET=F55hjj2FBPnnuW7nD80LjwS9sVYXm4fB
```
idanoo: I'm sure this is just for testing.. but I'd probably remove these from the PR

docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
       - external
       - internal
     ports:
-      - "8080:80"
+      - "8090:80"
```
idanoo:
```suggestion
      - "8080:80"
```

Probably want to keep this 8080

resources/views/auth/login.blade.php
```diff
@@ -7,6 +7,8 @@
             <div class="">
                 <div class="card-header bg-transparent p-3 text-center font-weight-bold h3">{{ __('Login') }}</div>
 
+		<a href="{{ route('oidc.signin') }}">Login with Keycloak</a>
```
idanoo: I assume will want check something like

`OIDC_ENABLED=true`

here so it's now showing to instances that don't use it

osresearch:
I've rebased on `dev`, although now I can't rebuild the docker image to test it.  My PHP know-how is really non-existant, so I'm at a loss as to why compose is throwing an error now about not finding the `oidc-client` library in the local directory, when it worked before.

```
Generating optimized autoload files
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi
In ProviderRepository.php line 208:
  Class 'GCS\OIDCClient\OIDCServiceProvider' not found  
```

Adding a `composer update` into the `contrib/docker/Dockerfile.apache` allowed the `docker-compose build` to finish, although now there are runtime errors at startup that I'm at a loss to debug.

```
app_1     | [2022-11-07 09:49:44] production.ERROR: Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php:184)
app_1     | [stacktrace]
app_1     | #0 /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php(184): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
```

~~Although this might not be my problem... a clean check out of `dev` 2bb27229f3f8c315d193010106fc328b6f3ac578 has the exact same error when run with `docker-compose build && docker-compose up`~~ I don't know how to `docker` or `php`... it looks like there is residual state from an old install in some docker volume that is causing this error.

osresearch:
IMG_0652

Some reworking to get things running on the latest checkout of `dev`, along with the patches to update the `docker-compose` environment to work with the new PHP version as well as to do all the initial setup.  The other big news is that these SSO patches work with the official Pixelfed mobile app (as of the current testflight version, not sure how to tell what that is)

I tried adding some config tests with the login screens, but it didn't work the way I had thought it would. I must be missing something in the way the template system works.

resources/views/auth/login.blade.php
osresearch: I've added a `config('oidc.enabled')` check in the two login files

.env.docker
osresearch: removed! I've also restored the `.env.docker` to its default (empty) form

verymilan:
any news on this? would be very nice to have...

ThisIsMissEm:
@osresearch I think this PR will conflict heavily with the recent "login with mastodon" support, so either needs to be rebased or closed.

oidc-client/src/Auth/OIDCUserProvider.php
```diff
@@ -0,0 +1,130 @@
...
+{
+    private function is_admin($user_info)
+    {
+	    if (!property_exists($user_info, "resource_access"))
+		    return false;
+	    if (!property_exists($user_info->resource_access, "pixelfed"))
+		    return false;
+	    if (!property_exists($user_info->resource_access->pixelfed, "roles"))
+		    return false;
+	    return in_array("admin", $user_info->resource_access->pixelfed->roles);
```
ThisIsMissEm, line 21: These are specific to how you configure keycloak, and can't be relied on for OIDC, the Mastodon PR had a similar issue too. (If you want a different provider to test with try Ory.sh)

ThisIsMissEm: That is, you can only really rely on the values in the ID Token or Access Token: 
- https://openid.net/specs/openid-connect-core-1_0.html#IDToken 
- https://openid.net/specs/openid-connect-core-1_0.html#Claims (standard claims)

dansup:
As @ThisIsMissEm mentioned, this will conflict with "sign-in with Mastodon" and upcoming 2FA refactor. I am going to rebase this into an optional package that can be installed along side pixelfed using composer and [laravel packages](https://laravel.com/docs/10.x/packages#introduction) to keep OIDC and SSO separate from the core project.

colega:
Can you point to the "Sign in with Mastodon" work? Issue or doc.

Edit: Google is my friend: https://pixelfed.blog/p/2023/feature/introducing-sign-in-with-mastodon-streamlining-sign-up-and-login-on-pixelfed

Not being able to define single identity for my Mastodon & Pixelfed users has been the biggest blocker to run Pixelfed so far (I don't want users to have different account names on them), so I was wondering if I could achieve that through OIDC.

PR #3452, opened 2022-05-14
Migrated to DATE_ATOM - DATE_ISO8601 is outdated

shleeable:
per docs https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.iso8601


dansup:
Fixed in #3499

PR #3480, opened 2022-05-17
Code cleaning. Unrequired breaks after returns

shleeable:


app/Http/Controllers/Api/AdminApiController.php
```diff
@@ -69,7 +69,7 @@ public function activity(Request $request)
     public function moderateStatus(Request $request)
     {
         abort(400, 'Unpublished API');
-        return;
+        
         $this->validate($request, [
```
shleeable: is all of this dead code?

dansup:
Removing `break` can make it harder to follow IMO

PR #3487, opened 2022-05-18
Update Helpers.php

shleeable:


app/Util/ActivityPub/Helpers.php
```diff
@@ -300,7 +300,7 @@ public static function statusFirstOrFetch($url, $replyTo = false)
 
 		$res = self::fetchFromUrl($url);
 
-		if(!$res || empty($res) || isset($res['error']) || !isset($res['@context']) ) {
+		if(isset($res['error']) || !isset($res['@context']) ) {
```
dansup, line 303: `$res` can be false and this wouldn't catch it

app/Util/ActivityPub/Helpers.php
```diff
@@ -495,7 +495,7 @@ public static function storeStatus($url, $profile, $activity)
 
 	public static function getSensitive($activity)
 	{
-		$id = isset($activity['id']) ? self::pluckval($activity['id']) : self::pluckval($url);
+		$id = isset($activity['id']) ? self::pluckval($activity['id']) : self::pluckval($activity['url']);
```
dansup, line 498: This does not check if `$activity['url']` isset

PR #3489, opened 2022-05-18
Update StatusEntityLexer.php

shleeable:
https://laravel.com/docs/5.2/helpers#method-str-slug


app/Jobs/StatusPipeline/StatusEntityLexer.php
```diff
@@ -102,7 +102,7 @@ public function storeHashtags()
 				continue;
 			}
 			DB::transaction(function () use ($status, $tag) {
-				$slug = str_slug($tag, '-', false);
+				$slug = str_slug($tag, '-');
```
dansup, line 105: We need the third parameter to force english slugs

Issue #3550, opened 2022-06-18
Username lenght limit


MrAn0nym:
Just a quick question since i wasn't able to find the answer online: Why does pixelfed limit the username length to 15 characters? Especially for migration from instagram it would imo make more sense to use the same rules as they do
https://github.com/pixelfed/pixelfed/blob/4f3a0d0ead98bdc3a0d7c7085463504009e84229/app/Http/Controllers/Auth/RegisterController.php#L68-L72

MrAn0nym:
Instagram seems to limit it to 30 characters, if you don't mind I'm going to create a pr tomorrow

MrAn0nym:
Any updates on this or the pr?

MrAn0nym:
Also another argument would be that mastodon has 30chars as well

McPringle:
I can't use the same username I have on Mastodon, PeerTube, YouTube, and Instagram here on Pixelfed. That is very annoying and preventing me from creating an account for my cycling activities. My username, which is known to the community (and I'm running a website using this as the domain, too), is 17 characters long and everywhere the same. I don't see a way to shorten it without loosing the meaning.

For me as a user, it is just annoying. For me as a software developer, I would suggest to go with the "unwritten standard" of 30 characters. That would be really great! Thank you very much!

FediVideos:
This might be good to extend to 30 now that there are so many people joining Pixelfed?

A lot of people and organisations have names longer than 15 characters.

PR #3554, opened 2022-06-20
30 character usernames

MrAn0nym:
Those would be all the changes required to change the username limit to 30 characters. 

Main reason for this change would be feature parity with Instagram but i personally experienced 15 characters being to short as well (issue #3550)


MrAn0nym:
Any updates?

MrAn0nym:
Delayed until the official release of Metro 2.0

dansup:
Closing for now, as previously discussed this is waiting on Metro v3.

McPringle:
I don't quite understand why this PR was closed. This usually means that it will not be merged. However, I could not find a reason for the rejection. It is mentioned that they are waiting for Metro v3. But that would be waiting and not rejecting. So I am confused about this.

Because I am also interested in adjusting the length of the username to that of Instagram, as I want to move my Instagram account to a Pixelfed instance. Of course, while keeping my username (17 characters), as I am known by it in the community and it cannot be shortened in a meaningful way without changing the meaning. As it is not possible to change usernames at a later date, this point is currently preventing me from migrating.

Since I believe that this is not only the case for me, but also for other Instagram users, an adjustment of the configuration would be very helpful.

Thank you very much for the Pixelfed software, keep up the great work!

Sythelux:
again really bad communication:

it was fixed in this commit: https://github.com/pixelfed/pixelfed/commit/207bb36c89f1ac08f0e62159a01dbd1ee67a356a

PR #3570, opened 2022-07-04
BUGFIX: Update Installer.php

shleeable:
Caught that the mitigation process reads the old config cache, and this resolves this issue.
image

The config needs to be refreshed before the mitigation is completed.

and split out the DB migration tasks and some artisan tasks


shleeable:
Actually skip this.

Issue #3193, opened 2022-01-28
Atom feed doesn't update when posting multiple images at once


tebjan:
Hello, my Atom feed doesn't update... My profile has two posts, but the Atom feed has only one entry. What could be the cause of this?

https://pixelfed.social/tebjan

https://pixelfed.social/users/tebjan.atom

EDIT: The issue seems to be a post with multiple images. I've just made a new one with just one image and that one is now visible in the feed.

raphraph:
Same on my feed.
It encourages in posting multiple (less valuable imho) posts for a set of pics instead of a single (more valuable) post in order to "feed" the feed...

aubguillemette:
Same here, would be nice if the Atom feed would include posts with multiple images.

niccokunzmann:
Searching...
Code:
https://github.com/pixelfed/pixelfed/blob/5d9b6863b66f092100380f88d00df65086e4a1d1/resources/views/atom/user.blade.php#L31

Generated code:

```
		<summary type="html">
		<![CDATA[
			<img id="rss_item_1" src="https://pix.toot.wales/storage/m/_v2/454898839954469424/3c6c7d1ab-ede048/odnDv854NNZE/G42QMeKxnWB0tJaDSRA8NNusyTGChcwSXnGk4ofv.jpg" alt="">
			<p style="padding:10px;">A little update: several outdoor signs for the &lt;a href=&quot;https://pix.toot.wales/discover/tags/INCC?src=hash&quot; title=&quot;#INCC&quot; class=&quot;u-url hashtag&quot; rel=&quot;external nofollow noopener&quot;&gt;#INCC&lt;/a&gt; are ready to be hung up. Made in the &lt;a href=&quot;https://pix.toot.wales/discover/tags/makerspace?src=hash&quot; title=&quot;#makerspace&quot; class=&quot;u-url hashtag&quot; rel=&quot;external nofollow noopener&quot;&gt;#makerspace&lt;/a&gt;!</p>
		  ]]>
		</summary>
```

What would the optimal feed look like ... one entry with several pictures? One entry per picture?

This claims that they should be embedded differently: https://perishablepress.com/feed-your-image-via-atom-or-rss/
```
<channel>
	<image>
		<url>http://domain.tld/path/site_logo.jpg</url> // *required! - the URL of the image
		<title>The alt attribute of the image</title>   // *required! - the image alt attribute
		<link>http://domain.tld/my_blog/</link>         // *required! - the URL of the website
[...]

PR #3627, opened 2022-08-20
display all photos in the atom feed

niccokunzmann:
until now, the atom feed only displays 10 images
this makes the atom feed display all photo feeds

related: https://github.com/pixelfed/pixelfed/issues/3193

Disclaimer: I am messing around a bit and I know it is usually worthless to just create a commit without a setup and running tests. BUT: at least we have a conversation!


dansup:
This could load hundreds or thousands of entries, which isn't ideal.

PR #3710, opened 2022-10-22
fix: add opt of forceScheme(https)

thisislawatts:
When running the application locally with
APP_URL=http://localhost it is unexpected for
all route URLs returned by `route('route-name')`
to use the https prefix.

Configuring SSL for your local environment should
not be a required step to development locally.

The new logic checks the configured URL for presence
of https:// OR that this the application is running
in a production environment.

I have not worked with Laravel before, so very open to pointers
for alternative places this could be handled. 🧠

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -39,7 +39,9 @@ class AppServiceProvider extends ServiceProvider
 	 */
 	public function boot()
 	{
-		URL::forceScheme('https');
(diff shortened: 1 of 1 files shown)


dansup:
@thisislawatts Thanks for the PR, this isn't the ideal way to disable it as there could be situations where your running in production and still want to disable it so I'll merge this and update it a bit to use the `FORCE_HTTPS_URLS` .env variable (enabled by default).

PR #3725, opened 2022-11-06
Remove erroneous HTML-escaping in Danish translation

ghost:
The Danish translations have HTML entities, e.g. `S&oslash;g` instead of `Søg`. Literally none of the other translations have this. Combine this with the fact that I was seeing `S&oslash;g` via the UI, and I assume that this is a bug, which this PR therefore fixes.


dansup:
Please send translation updates via https://translate.pixelfed.org/

PR #2137, opened 2020-04-21
docker: reduce layers, remove unused packages

Strubbl:
this is just a draft, open for discussion

Issue #3597, opened 2022-07-31
Docker-compose clean install doesn't function


ggrensteiner:
On install of a new docker-compose instance using the supplied docker-compose.yml does not give a functional pixelfed_app container. It looks like the laravel code is broken on some kind of missing driver declaration on the db initialization.

```
+ cp -r storage.skel/app storage.skel/debugbar storage.skel/framework storage.skel/logs storage.skel/purify storage/
+ php /wait-for-db.php
Database is up
+ [[ ! -e storage/.docker.init ]]
+ echo 'Fresh installation, initializing database...'
+ gosu www-data php artisan key:generate
Fresh installation, initializing database...
[2022-07-31 13:08:29] production.ERROR: Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php:184)
```

The db container is online and working, connectivity works between the app and the db using the .env.production vars that I have entered.

leafonthewind:
Thanks for this!  I was upgrading from 0.11.3 to 0.11.4 today and ran into this same error (pasted below for completeness).  This change in `.env.docker` solved it.

Much appreciated!

```bash
[2022-10-10 13:52:23] production.ERROR: Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php:184)
[stacktrace]
#0 /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php(184): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
#1 /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php(169): Illuminate\\Broadcasting\\BroadcastManager->resolve()
#2 /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php(158): Illuminate\\Broadcasting\\BroadcastManager->get()
#3 /var/www/vendor/laravel/framework/src/Illuminate/Broadcasting/BroadcastManager.php(384): Illuminate\\Broadcasting\\BroadcastManager->driver()
[...]

PR #3735, opened 2022-11-09
docker: Remove bootstrap volume and streamline startup

osresearch:
This patch modifies the `start.apache.sh` script to perform the initial setup on first boot of the container so that the database and all of the various keys are applied.  I'm not positive these are the right steps in the right order, although it allows me to bring my pixelfed server online.

The patch also moves the storage, db, and redis volumes into a local `data` directory for easier management and removes the bootstrap volume since it did not seem necessary for normal operation.  Was there a reason for the separate `bootstrap` volume that I missed in [the `cp -r storage.skel` discussion](https://github.com/pixelfed/pixelfed/pull/2137#discussion_r434468862)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ yarn-error.log
 .git-credentials
 /.composer/
 /nginx.conf
+/data
--- a/contrib/docker/Dockerfile.apache
+++ b/contrib/docker/Dockerfile.apache
@@ -1,10 +1,10 @@
-FROM php:7.4-apache-bullseye
+FROM php:8.1-apache
 
 # Use the default production configuration
 COPY contrib/docker/php.production.ini "$PHP_INI_DIR/php.ini"
 
 # Install Composer
-ENV COMPOSER_VERSION=2.1.14 \
+ENV COMPOSER_VERSION=2.4.4 \
     COMPOSER_HOME=/var/www/.composer \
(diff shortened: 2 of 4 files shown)


albjeremias:
let me test this :)

albjeremias:
didn't work.. this worked:
https://github.com/pixelfed/pixelfed/issues/3597#issuecomment-1319287683

I removed the app-storage as a volume on docker-compose

and re-run so that bootstrap files are generated again?!
`$ docker-compose exec app composer install --prefer-dist --no-interaction --no-ansi --optimize-autoloader`

osresearch:
I've rebased to dev and merged the docker fixes from #3740, as well as added a comment about the documentation on the one-time startup tasks that are performed.

Strubbl:
if i remember correctly, this needs to be rebased against staging. After the PR is merged into staging, there is a merge to dev done manually by dansup.

dansup:
> if i remember correctly, this needs to be rebased against staging. After the PR is merged into staging, there is a merge to dev done manually by dansup.

This is correct, please rebase against the staging branch!

bplein:
Personal opinion: it’s better to keep all of the underlying services in their own volume. There are many reasons for this, including the ability to easily place different things in different performance or price storage.

dansup:
Superseded by #4844

PR #3737, opened 2022-11-10
Update Dockerfile.apache

rvleij:
The old PHP version prevents building the container image. It builds fine when "upgrading" towards 8.1, hence this change.


harmtemolder:

dansup:
Fixed in #4050

PR #3740, opened 2022-11-10
Updates Dockerfile.apache to PHP 8.1 and Composer 2.4.4

der-On:


rvleij (approved):
I did the same PR (but without the Composer update), also verified it worked (the PHP 8.1 update), so I approve1

deftdev (approved):

rvleij (approved):

osresearch:
This patch worked for me for a clean checkout, after removing any old docker volumes.

M-D-M (approved):
I can confirm that this builds on docker 20.10.17 on Ubuntu 20.04.

M-D-M:
Speaking on launching Pixelfed via Docker, how is everyone handling HTTPS?  That's the biggest problem I'm having right now.

der-On:
> Speaking on launching Pixelfed via Docker, how is everyone handling HTTPS? That's the biggest problem I'm having right now.

I usually use nginx as reverse proxy for my docker based setups. Inside docker everything runs with http. Nginx then uses letsencrypt signed certificates. This allows me to test everything locally and prevents complex https setups inside the containers.

M-D-M:
Yep -- I'm just starting to figure out how to run a reverse proxy with nginx now.  I wonder if it would be worth building that into the docker-compose file?

der-On:
@M-D-M I don't think it would make much sense. There are numerous docker images/setups for reverse proxies with nginx using let's encrypt and as every hosting situation is different there would not be a one-fits-all solution.

osresearch:
nginx setup is probably out of scope for this pull request (which really needs to be merged!). It might be worth creating a new PR with some details on how to setup caching so that pixelfed can better handle the thundering herd of a popular boost, as well as simplifying the certbot integration.

mattlqx:
re: nginx. I created this to get an image to run. https://github.com/mattlqx/docker-pixelfed

harmtemolder:

dansup:
Fixed in #4050

PR #3779, opened 2022-11-16
add FORCE_HTTPS config

OriHoch:
currently app force https urls, this configuration allows to optionally disable this and serve http urls (for development)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,6 +22,7 @@ APP_DOMAIN="localhost"
 ADMIN_DOMAIN="localhost"
 SESSION_DOMAIN="localhost"
 TRUST_PROXIES="*"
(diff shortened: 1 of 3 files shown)


albjeremias:
i'd rather have this config as optional, so if you don't put the the FORCE_HTTPS it would force it anyway.
The issue is that the .env is becoming way too big and complicated.

NO_HTTPS=true ?

HaraldWalker:
Using this change and after I also changed HTTPS Only Cookies to `'secure' => false` in `config/session.php` I was finally able to login on my local VM where I can now reproduce a production bug.

dansup:
Added in a31bdec7, using a slightly different .env variable!

PR #3800, opened 2022-11-18
Add Norwegian Bokmål (norsk bokmål) translations

Cyrus-Harding:
Added translations for Norwegian Bokmål (norsk bokmål).

Added symlink from nb to no.

For those who don't know, Norwegian has two written languages: Bokmål and Nynorsk.

I don't know if I did it correctly, I tried to do the same as in #40, but when I try it on my server, only the footer in the settings seems to use the localization.

image

The Danish translation doesn't seem to work either.


dansup:
Please send translation changes via https://translate.pixelfed.org/

Issue #1621, opened 2019-08-19
Allow registering without setting display name


poperigby:
Is it intended that you have to have a name for your account? What's the distinction between a username and a name? Is the name supposed to be your IRL name? If so, that doesn't seem like a good idea to make that mandatory.

dansup:
@PopeRigby The name field is optional but the register form does not reflect that, will fix!

poperigby:
You should also add a placeholder of "Optional"

dansup:
Fixed in the latest commits, we added a `required` attribute to the `name` field and also updated the backend to accept empty or null `name` values!

PR #3812, opened 2022-11-20
Update register.blade.php

### LINKED ISSUE #3811: Update register page to match login page
My changes to the login page was accepted awhile back so will submit one for register page

Here is what it looks like currently

pixelfedregistercurrent

okpierre:
Minor adjustments to fix https://github.com/pixelfed/pixelfed/issues/3811

See screen shot

pixelfedregisterpage

### DIFF
--- a/resources/views/auth/register.blade.php
+++ b/resources/views/auth/register.blade.php
@@ -4,8 +4,8 @@
 <div class="container mt-4">
     <div class="row justify-content-center">
         <div class="col-lg-5">
-            <div class="card">
-                <div class="card-header bg-white p-3 text-center font-weight-bold">{{ __('Register a new account') }}</div>
+            <div class="">
+                <div class="card-header bg-transparent p-3 text-center font-weight-bold h3">{{ __('Register a new account') }}</div>
 
                 <div class="card-body">
                     <form method="POST" action="{{ route('register') }}" class="px-md-3">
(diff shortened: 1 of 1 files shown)


dansup:
Please rebase against the staging branch!

okpierre:
@dansup done

mryoaz:
@okpierre Maybe an idea to also fix issue #1621 in this PR?
Or must that be done in an other PR?

svenluijten:
Removing the `<label>` elements entirely is a pretty big accessibility faux-pas. The labels didn't have `for` attributes to begin with, but they absolutely deserve to be there, even if only for screen-reader software. 

An alternative would be to add [`aria-label` attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) to all of the `<input>` elements so that this page can still be used by people using assistive technology to use Pixelfed.

okpierre:
@svenluijten Thanks for your input. Updated with aria-label. Let me know if I missed anything.

mitexleo:
I think old one looks better...

svenluijten:
@okpierre wouldn't it make sense to use the translation strings (`__('...')`) for the labels? That's what the placeholders are using. Also, the `<button>` element doesn't _need_ an `aria-label` attribute. From the page I linked above:

> By default, a button's accessible name is the content between the opening and closing `<button>` tags, an image's accessible name is the content of its `alt` attribute, and a form input's accessible name is the content of the associated `<label>` element.

As an aside, I've submitted a PR to fix this in all authentication views: https://github.com/pixelfed/pixelfed/pull/3980.

ThisIsMissEm:
@dansup I think you're redoing this screen or have redone it, so perhaps this needs to be closed?

dansup:
> @dansup I think you're redoing this screen or have redone it, so perhaps this needs to be closed?

Yes, I'm working on a whole new registration flow which isn't compatible with these changes, we appreciate your contribution though!

Issue #2712, opened 2021-04-04
Bug: "Invite someone" button is not displayed when there is at least one invite


inga-lovinde:
In [resources/views/settings/invites/home.blade.php](https://github.com/pixelfed/pixelfed/blob/dev/resources/views/settings/invites/home.blade.php):

* If there is at least one invite, list of invites is displayed;
* If there is no invites, a message "You haven't invited anyone yet" along with the number of invites left and with an "Invite someone" button is displayed.

What was probably intended is:

* If there is at least one invite, display list of invites;
* If there is no invites, display a message "You haven't invited anyone yet";
* In any case, display the number of invites left and an "Invite someone" button.

ghost:
Is there any known workaround to this issue? Or a fix that I somehow missed?

albjeremias:
i confirm. and also is not possible to remove a previous invitation

stevenroose:
This is a really big bug.. It means I can't invite a second person before the first person has accepted :sweat_smile:

stevenroose:
@ar-i @albjeremias a temporary workaround is to go to the "create invite" page manually:
`https://<domain>/settings/invites/create`

PR #3817, opened 2022-11-21
fix: when one invitation is send, cant send no more

albjeremias:
fixes: 
https://github.com/pixelfed/pixelfed/issues/2712


dansup:
User invites have not been released yet and are being refactored, I appreciate the sentiment though!

PR #3823, opened 2022-11-23
Enhance translation: Use phrase instead of single words

juengling:
I recommend to use the complete phrase "followed you" here instead of the single words "followed" and "you".


dansup:
Please send translation changes via https://translate.pixelfed.org/

PR #3825, opened 2022-11-24
fixes the problem with the creation of directories

goxonline:
the problem occurred when creating the directories, the permissions were not set correctly and prevented the web user from reading the directory.

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -46,6 +46,16 @@ return [
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
+            'permissions' => [
+                'file' => [
+                   'public' => 0644,
+                   'private' => 0644,
+                ],
+                'dir' => [
+                    'public' => 0755,
+                    'private' => 0755,
+                ],
+            ],
         ],
 
         'public' => [


config/filesystems.php
```diff
@@ -46,6 +46,16 @@
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
+            'permissions' => [
+                'file' => [
+                   'public' => 0644,
+                   'private' => 0644,
```
jemayer, line 52: Are you sure setting "private" permissions to be accessible by the public is the right way to go? I don't know where and how these are used with Pixelfed, but it doesn't look like a good pattern. (Same goes for the directory permissions.)

goxonline: Good morning, 
Yes I have checked it. But I think you can remove the "public" parameter, since you are only using/respecting the "private". 
I have left both, in case I missed something.
Regards

goxonline: Also, as in my case I am running PHP under PHP-FPM, the user should always create the files and directories with permissions 0644/0755 respectively, otherwise the user running the WEB service will not be able to access these files generating a 403 error.

creideiki: > Also, as in my case I am running PHP under PHP-FPM, the user should always create the files and directories with permissions 0644/0755 respectively, otherwise the user running the WEB service will not be able to access these files generating a 403 error.

I am also attempting to run Pixelfed like this, and this patch solves the same problem for me. However, I don't know enough about PHP or Pixelfed to opine on whether this opens any security vulnerabilities.

goxonline: No security holes opens with this fix. The permissions 0755 and 644 are the commons use in web content.

dannekrose:
I apologize for my ignorance, but after I apply the patch to the local file, how do I get it to "apply" or take?  I did it, but I still have the same folder permissions where new folders are being created with the wrong permissions.  I've done full box reboots as well.

creideiki:
> I apologize for my ignorance, but after I apply the patch to the local file, how do I get it to "apply" or take?

You must update the caches. See https://docs.pixelfed.org/running-pixelfed/installation/#one-time-setup-tasks :

> Routes should be cached whenever the source code changes or whenever you change routes:
>
> `php artisan route:cache`
> `php artisan view:cache`
>
> Every time you edit your .env file, you must run this command to have the changes take effect:
>
> `php artisan config:cache`

HaraldWalker:
Ran into the same problem (Ubuntu + NGINX) with a new installation and local storage. The fix of the PR did not solve it for me. For now I resolved it by switching to `FILESYSTEM_DRIVER=public`.

escuco:
Hello and thanks for the patch!

I've the same problem with the picture upload, applied the patch and updated the caches - but still the picture upload no longer works.

Do you have any further hints for me?

HaraldWalker:
@escuco 
You might have folders with the old invalid permissions preventing nginx from accessing the folders from before you applied the patch. You can try to fix those permissions or if there is not much/anything uploaded yet, delete them. In my case I deleted `/home/pixelfed/pixelfed/storage/app/m` and since then it is working fine with this patch.

escuco:
Hi @HaraldWalker 

many thanks, that did the trick.

escuco (approved):

config/filesystems.php
ltning: Everything else - including the files - are created with these perms already, so it makes no sense for the parent directories not to be. Current behaviour breaks any attempt at separating webserver from php-fpm.

ltning (approved):

ltning:
I would really like to see this merged, as mentioned in my attempted "approve" comment.

jemayer (approved):
As even the installation docs include the following snippet for setting correct permissions, I see no reason no to merge this one.

```
sudo find . -type d -exec chmod 755 {} \; # set all directories to rwx by user/group
sudo find . -type f -exec chmod 644 {} \; # set all files to rw by user/group
```

ltning:
There seems to be some confusion about what the actual problem is. While chmod/chown helps some things, there's a very specific problem with thumb/preview generation: The directory created for the previews for an individual post is created with 0700 permissions, despite everything else having 0755 or whatever. This means that if the webserver - nginx in my case - runs as a different user than php-fpm (which is good security practice, for one), it cannot read the previews.
The previews *themselves* are readable, but without (at the very least) the execute flag set on the directory, nginx cannot traverse it to access the files.

Example directory from before the patch:
`drwx------  2 www  www  2 Jan  1 13:36 /usr/local/www/pixelfed/public/storage/m/_v2/515139097376215041/3079cad20-917577/rfARreMan5Te`

After the patch:
`drwxr-xr-x  2 www  www  4 Jan  1 15:06 /usr/local/www/pixelfed/public/storage/m/_v2/515139097376215041/3079cad20-917577/IUNoZM2PKKKP/
`

Now if nginx, running as `nobody` which is member of the `www` group, it can read the files (but not modify them). Without the patch it could not even traverse the directory.

MetroWind:
Also had this problem. I was able to get it working with this change:
```
--- filesystems.php	2023-02-26 13:57:08.030035215 -0800
+++ filesystems.php.new	2023-02-26 13:58:28.653331554 -0800
@@ -46,6 +46,12 @@
         'local' => [
             'driver' => 'local',
             'root'   => storage_path('app'),
+            'permissions' => [
+                'dir' => [
+                    'public' => 0755,
+                    'private' => 0755,
+                ],
+            ],
         ],

         'public' => [
```

mitexleo:
Having the same problem 😕

PR #3887, opened 2022-12-03
Update email and password views with the settings template

mryoaz:
This pull request updates the email and password views, in account settings, to also use settings template. Just like all other views in the account settings. 
It makes alle the account settings pages consistent and you won't need to manually navigate back.

Old email settings page
New email settings page
Old password settings page
New password settings page


dansup:
This was an intentional change to support webview clients, thanks anyways!

Issue #3857, opened 2022-12-01
Unable to upload JPG image as story - getting "invalid media type" error


zcraber:
I'm unable to upload [this ](https://user-images.githubusercontent.com/61133303/205126991-16532a62-f760-463b-85f0-7a8793572e50.jpg)JPG image as a story. Getting `Invalid media type` error.
![image](https://user-images.githubusercontent.com/61133303/205127568-e27f1229-3b56-44fd-94ad-67f53630d9d7.png)

[pixey.org](https://pixey.org/) server. Pixelfed v0.11.4.

vanlueckn:
Can confirm it, i tried to upload your jpg on my instance and get the same error.

vanlueckn:
Your Image is a webp image. MIME Type is `image/webp`.

I have webp images activated in my instance but the story upload does not accept webp images anyway.

zcraber:
> Your Image is a webp image. MIME Type is `image/webp`.
> 
> I have webp images activated in my instance but the story upload does not accept webp images anyway.

Oh, ok. I'll close it then. WebP is not supported for stories, right? I could upload it as a post without issues. #2791

vanlueckn:
I created a pull request with code changes allowing webp stories: #3888

PR #3888, opened 2022-12-03
Allow webp images in stories

vanlueckn:
Refers to #3857 

I have added webp as a mime type to the stories. I have tested in my local development instance the upload of WEBP images afterwards - It works so far without problems.

Test Story with webp image from #3857:
brave_DeRHaJZe9j


dansup:
While we do plan on extending Story support to other formats, it requires additional work regarding federation validation and the transcoding pipeline. This PR doesn't cover those so I will be closing this, I appreciate your work on this and will ping you once we ship webp support in Stories!

PR #3894, opened 2022-12-04
Add a command to import emoji archives to pixelfed

vanlueckn:
### Summary:
This pull request adds a new command to the pixelfed CLI that allows users to import emojis from a `.tar.gz`  archive. The command takes a file path as an argument and adds the emoji to the databaset.

### Description:
The new command is called `php artisan import:emoji` and it is added to the `pixelfed` CLI. To use the command, the user must provide a file path to a file containing the emojis they want to import.

The command first extracts the archive and then iterates over all files. All emojis are then added to the database.

### Signature
The signature of the command:
````
import:emojis
{path : Path to a tar.gz archive with the emojis}
{--prefix : Define a prefix for the emjoi shortcode}
{--suffix : Define a suffix for the emjoi shortcode}
{--overwrite : Overwrite existing emojis}
{--disabled : Import all emojis as disabled}
````

### Motivation
I find the Mastodon feature to import whole emoji archives very handy. Currently in Pixelfed you have to create each emoji individually with a lot of effort.

### DIFF
--- a/app/Console/Commands/ImportEmojis.php
+++ b/app/Console/Commands/ImportEmojis.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\CustomEmoji;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Storage;
+
+class ImportEmojis extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'import:emojis
+                            {path : Path to a tar.gz archive with the emojis}
+                            {--prefix : Define a prefix for the emjoi shortcode}
+                            {--suffix : Define a suffix for the emjoi shortcode}
+                            {--overwrite : Overwrite existing emojis}
+                            {--disabled : Import all emojis as disabled}';
+
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Import emojis to the database';
+
+    /**
+     * Execute the console command.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/Console/Commands/ImportEmojis.php [hunk 1]
```diff
@@ -0,0 +1,110 @@
...
+
+        foreach (new \RecursiveIteratorIterator($tar) as $entry) {
+            $this->line("Processing {$entry->getFilename()}");
+            if (!$entry->isFile() || !$this->isImage($entry)) {
+                $failed++;
+                continue;
+            }
+
+            $filename = pathinfo($entry->getFilename(), PATHINFO_FILENAME);
+            $extension = pathinfo($entry->getFilename(), PATHINFO_EXTENSION);
```
elza3ym, line 61: @vanlueckn Possible security issue, Please add validation on the extension and mime-type to be a valid image/emoji.

vanlueckn: Thank you for your review, i will push the fix in a few hours.

vanlueckn: I pushed a fix

app/Console/Commands/ImportEmojis.php
```diff
@@ -0,0 +1,118 @@
...
+
+            // Skip macOS shadow files
+            if (str_starts_with($filename, '._')) {
+                continue;
+            }
```
svenluijten, line 66: Won't line 55 (checking if it is a file, and checking if it is an image) already handle these cases?

dansup (approved):

dansup:
Thanks!

PR #3980, opened 2022-12-20
Make authentication views accessible

svenluijten:
The inputs on the authentication views didn't have proper labelling, making them almost unreadable for screen readers. This PR fixes that by adding [the `aria-label` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) to all relevant input fields in the `auth` folder. 

I've also taken the liberty to increase the legibility of the code by putting every attribute on its own line for the input fields. It was getting kind of long 🙂

### DIFF
--- a/resources/views/auth/checkpoint.blade.php
+++ b/resources/views/auth/checkpoint.blade.php
@@ -19,7 +19,7 @@
                         <div class="form-group row">
 
                             <div class="col-md-12">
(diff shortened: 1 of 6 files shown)


intentionally-left-nil (approved):

svenluijten:
Closing due to inactivity. If this is still (or eventually becomes) a priority, I'll be happy to take another look.

PR #3984, opened 2022-12-21
Remove unnecessary empty constructors

svenluijten:
We can remove constructors that _only_ contain `parent::__construct();` since these will fall back to their parent class automatically. This cleans up the code and makes it easier to read when first opening the file(s).

### DIFF
--- a/app/Console/Commands/AvatarDefaultMigration.php
+++ b/app/Console/Commands/AvatarDefaultMigration.php
@@ -24,16 +24,6 @@ class AvatarDefaultMigration extends Command
     protected $description = 'Replace old svg identicon avatars with default png avatar';
 
     /**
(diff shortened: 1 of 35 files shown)


svenluijten:
Sorry about the (unrelated) whitespace changes, my editor does this automatically when changing files. It is also compliant with the `.editorconfig` file present in this repository, so I'm not actually sure where these came from to begin with...

dansup:
Sorry, this PR is too big and consists of mostly superficial code style changes.

svenluijten:
Makes me wonder why you have an `.editorconfig` file that applies these changes to begin with then 🙂 Like I said in my comment, the (superficial) code style changes were applied by my editor picking up a config file you distribute with this source code.

PR #3989, opened 2022-12-22
Add .ddev configuration for local development

weitzman:
See [DDEV](https://ddev.readthedocs.io/en/stable/).

I think these small files will help jumpstart local development for users who care to try it. With this PR, `ddev start` sets up a functional site. Database, email, and redis are all configured. Emails may be viewed via DDev's Mailhog service. For local developers, this should be considerably easier than our (non-functional) docker-compose file.


.ddev/config.yaml
```diff
@@ -0,0 +1,33 @@
+name: pixelfed
```
rfay:
```suggestion
```

Recommend leaving this out, so it will just take the directory name.

rfay (commented):
Minor suggestion about leaving out the `name:`

.ddev/config.yaml
weitzman: Done

dansup (approved):

PR #4009, opened 2022-12-24
Docker hub push workflow

Murazaki:
Here is an image build workflow based on @kovah's fork and the mastodon's one.

Works right now on my fork, and you can pull images at https://hub.docker.com/r/murazaki/pixelfed

I'll try to update it in the meantime.


Murazaki:
Gonna move this into a new branch so that the fork can be automatically updated and built meanwhile.

PR #4047, opened 2022-12-31
Staging

dansup:


juliloving:
"Juno" was left out and is still trademarked

Src: https://about.meta.com/brand/resources/meta/our-trademarks

PR #4050, opened 2022-12-31
Build CI and docker updates for automatic image push to Docker Hub

Murazaki:
Cleaned up branch.
Fork is now delivering images at Murazaki/pixelfed (and automatically rebasing itself, hence the separate branch with different image naming).
I hope you'll find it useful.

PR #4065, opened 2023-01-06
 Update Dockerfile.apache 

davidestf:
Update image from php:7.4 to php:8.1-apache-bullseye


harmtemolder:

dansup:
Fixed in #4050

PR #4074, opened 2023-01-09
Clean up and improve Docker setup

harmtemolder:
Following #4050 I've gone ahead and cleaned up the Docker setup even further:

- The `.env.example` and `.env.docker` files now contain all possible variables, commented
- `Dockerfile.apache` is split in a couple more `RUN`s so that rebuilding the image locally can reuse cached steps for each
- `start.apache.sh` now initializes a new install and runs database migrations when needed
- Added a `worker.apache.sh` that will check if the new install is initialized and if there are any pending migrations before starting
- Added `depends_on` conditions to `docker-compose.yml` to make sure MariaDB is fully up before starting Pixelfed
- Added names to all services and networks for easier identification

@Murazaki and @Kovah Seeing that you were actively working on the Docker setup before, what do you think of all of these improvements?


harmtemolder:
Also updated `.editorconfig` to match the code style I encountered in all of these files. And added to `.dockerignore`

Kovah:
The changes look reasonable. But are you also planning to automatically build a Docker image that is available publicly? I see that the docker-compose file uses `pixelfed/pixelfed:latest` but there is no such image on Docker Hub.

Murazaki:
That's a lot of changes.

I like the idea of adding a migration step to the start script, if it won't risk breaking the user data.

I don't think though the docker rewrite is ok : we need to do an apt cleanup in the same command as the apt install if we want to limit the size of the cached steps.

Murazaki:
> The changes look reasonable. But are you also planning to automatically build a Docker image that is available publicly? I see that the docker-compose file uses `pixelfed/pixelfed:latest` but there is no such image on Docker Hub.

This comes from my PR, we're waiting for them to activate the github actions on this repo though.

harmtemolder:
@Murazaki Good point about the cleanup. I moved those to their respective steps

I'm now figuring out what happens here:

```dockerfile
# for detail why storage is copied this way, pls refer to https://github.com/pixelfed/pixelfed/pull/2137#discussion_r434468862
RUN cp -r storage storage.skel \
    && composer install --prefer-dist --no-interaction --no-ansi --optimize-autoloader \
    && rm -rf html && ln -s public html \
    && chown -R www-data:www-data /var/www
```

And if that can (partly) be moved to before `COPY . /var/www/`

harmtemolder:
Never mind, the `composer install` depends on all kinds of things that are only available after the `COPY`, so I'll leave that bit as is

Everything else seems to be working smoothly, both on my local machine and on my server

dansup (changes requested):
Why did you comment out all the `.env.example` variables? Please uncomment all variables that currently exist in `.env.example`

harmtemolder:
Hi @dansup, thanks for checking the PR. `.env.example` now contains all the default values used by the code, which is why I commented all of them: this way a new user would see what the default is, and can uncomment and change it if they need anything else. And if they don't change it, the default will still be used. Does that make sense?

Murazaki:
Honestly that's a lot of changes with a lot of opinionated views on linting, dockerfiles and docker-compose structure and also docker apps workflow. There's a lot to explain in there, most projects don't implement half of the changes you added in the docker-compose for instance and they still run fine.

I would try to approve the entrypoints first, if possible, with limited changes to docker compose, then maybe propose the rest in another PR.

harmtemolder:
@Murazaki Good idea, I'll start splitting this PR up.

Regarding the linting: the changes to `.editorconfig` actually follow what I encountered in each file type, so that I could lint while working without messing anything up.

Strubbl:
I agree, it would be nice if we can progress your cleanup and improvement steps to the docker setup in multiple smaller pull requests with less lines changed each. This might also help to get things merged easily.

harmtemolder:
I've split up the changes into these four PRs:

- #4184 
- #4185 
- #4186 
- #4187 

So I'll close this one

PR #4098, opened 2023-01-16
Suggested weight optimization for the SVG logo

nhoizey:
I ran the logo in [SVGOMG](https://jakearchibald.github.io/svgomg/), a tool for optimizing SVG images, and it saves 54 % of the weight without losing anything visually.

I kept the pretty markup, even if minification would have saved a few more bytes.


dansup:
Thanks, was able to shave an additional 9.4% by stripping whitespace!

nhoizey:
@dansup great! 👍

PR #4171, opened 2023-02-12
Publish docker image to GitHub registry

jgillich:
It appears that the image is currently not being published to Docker Hub, perhaps there is some setup missing.

This changes the workflow to publish to GitHub registry, which requires no setup and is better integrated with GitHub.

Sample builds are available on my fork: https://github.com/jgillich/pixelfed/pkgs/container/pixelfed


dansup:
Superseded by #4844

PR #4172, opened 2023-02-12
dockerfile: install all db drivers

jgillich:
Updates the Dockerfile to include Postgres and SQLite support. I feel like these are popular options that should be included by default. The only real downside is that the image size increases by a little bit.


intentionally-left-nil (approved):
Yeah that's one of the things I have to do in my fork:
https://github.com/intentionally-left-nil/pixelfed/blob/main/patches/0002-Add-postgres-and-sqlite-to-the-docker-images.patch

so it would be great to include it for everyone

dansup:
Superseded by #4844

Issue #3045, opened 2021-12-15
Class 'BeyondCode\SelfDiagnosis\SelfDiagnosisServiceProvider' not found


Strubbl:
I get an error when running `php artisan migrate --force` on the latest dev commit c580c3dc5efd202d1bf7fff00cae6b2a9459e731

```
php artisan migrate --force
[2021-12-15 10:06:33] production.ERROR: Class 'BeyondCode\SelfDiagnosis\SelfDiagnosisServiceProvider' not found {"exception":"[object] (Error(code: 0): Class 'BeyondCode\\SelfDiagnosis\\SelfDiagnosisServiceProvider' not found at /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:735)
[stacktrace]
#0 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(669): Illuminate\\Foundation\\Application->resolveProvider()
#1 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/ProviderRepository.php(75): Illuminate\\Foundation\\Application->register()
#2 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(649): Illuminate\\Foundation\\ProviderRepository->load()
#3 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/RegisterProviders.php(17): Illuminate\\Foundation\\Application->registerConfiguredProviders()
#4 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(237): Illuminate\\Foundation\\Bootstrap\\RegisterProviders->bootstrap()
#5 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Console/Kernel.php(310): Illuminate\\Foundation\\Application->bootstrapWith()
#6 /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Console/Kernel.php(127): Illuminate\\Foundation\\Console\\Kernel->bootstrap()
#7 /var/www/artisan(37): Illuminate\\Foundation\\Console\\Kernel->handle()
#8 {main}
"} 

In Application.php line 735:
                                                                           
  Class 'BeyondCode\SelfDiagnosis\SelfDiagnosisServiceProvider' not found
```

Strubbl:
The issue is still reproducible with dev commit 5442e232dc8d2255393c863ed7199d93b4b87ce9

Strubbl:
[...]

PR #4184, opened 2023-02-15
Improve Docker entrypoints

harmtemolder:
Splitting up #4074. This PR only includes the improvements to the Docker entrypoints:

- `start.apache.sh` now initializes a new install and runs database migrations when needed
- Added a `worker.apache.sh` that will check if the new install is initialized and if there are any pending migrations before starting

### DIFF
--- a/contrib/docker/start.apache.sh
+++ b/contrib/docker/start.apache.sh
@@ -1,15 +1,39 @@
 #!/bin/bash
 
+# We should already be in /var/www, but just to be explicit
+cd /var/www || exit
+
 # Create the storage tree if needed and fix permissions
 cp -r storage.skel/* storage/
 chown -R www-data:www-data storage/ bootstrap/
 
-# Refresh the environment
-php artisan storage:link
-php artisan horizon:publish
-php artisan route:cache
-php artisan view:cache
-php artisan config:cache
+if [[ ! -e storage/.docker.init ]]; then
+    echo "Fresh installation, initializing database..."
+    chown www-data:www-data .env &&
+        gosu www-data php artisan key:generate &&
+        gosu www-data php artisan migrate:fresh --force &&
+        gosu www-data php artisan passport:keys &&
+        echo "done" >storage/.docker.init
+fi
 
-# Finally run Apache
-apache2-foreground
+# Refresh the environment
+gosu www-data php artisan storage:link
+gosu www-data php artisan horizon:publish
+gosu www-data php artisan config:cache
+# gosu www-data php artisan cache:clear
+gosu www-data php artisan route:cache
+gosu www-data php artisan view:cache
+
+# Check for migrations
+gosu www-data php artisan migrate:status | grep No && migrations=yes || migrations=no
+
+if [ "$migrations" = "yes" ]; then
+    echo "Found outstanding migrations, running those..."
+    gosu www-data php artisan migrate --force
+fi
+
+# Create instance actor
+gosu www-data php artisan instance:actor
+
+# Finally run Apache, passing along any parameters from `CMD`
+dumb-init apache2-foreground "$@"
--- a/contrib/docker/Dockerfile.apache
+++ b/contrib/docker/Dockerfile.apache
@@ -97,4 +97,4 @@ RUN php artisan horizon:publish
 
 VOLUME /var/www/storage /var/www/bootstrap
 
-CMD ["/var/www/contrib/docker/start.apache.sh"]
+ENTRYPOINT ["/var/www/contrib/docker/start.apache.sh"]
--- a/contrib/docker/worker.apache.sh
+++ b/contrib/docker/worker.apache.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# We should already be in /var/www, but just to be explicit
+cd /var/www || exit
+
+if [[ ! -e storage/.docker.init ]]; then
+    echo "Database is not initialized yet, exiting..."
+    sleep 5
+    exit 1
+fi
+
+# Check for migrations
+gosu www-data php artisan migrate:status | grep No && migrations=yes || migrations=no
+
+if [ "$migrations" = "yes" ]; then
+    echo "Found outstanding migrations, exiting..."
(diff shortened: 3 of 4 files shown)


Murazaki:
This looks very promising and might help us streamline docker updates. Thanks.

Strubbl:
In my update script i also have a line to delete some cache files: `rm /data/pixelfed/bootstrap/cache/*`. I introduced this due to #3045. Does it make sense to add it here to the "Refresh the environment" section, too?

contrib/docker/start.apache.sh
```diff
@@ -1,15 +1,39 @@
...
-apache2-foreground
+gosu www-data php artisan storage:link
+gosu www-data php artisan horizon:publish
+gosu www-data php artisan config:cache
+# gosu www-data php artisan cache:clear
+gosu www-data php artisan route:cache
+gosu www-data php artisan view:cache
+
+# Check for migrations
+gosu www-data php artisan migrate:status | grep No && migrations=yes || migrations=no
```
Strubbl, line 28: This looks like it will break easily, because we parse the output of  the migrate:status command. I mean, is it ensured, that the migration names are always lowercase, so that this grep does not falsely finds a match?
I just had a short look into https://laravel.com/docs/master/migrations. But there is no more robust way to get the migration status, or?

Strubbl (approved):

harmtemolder:
@Strubbl I'm not really familiar with Laravel, so I don't actually know what that line does, or what the issue in #3045 was. If I'm reading correctly this line would be added like this:

```shell
if [ "$migrations" = "yes" ]; then
    echo "Found outstanding migrations, running those..."
    rm /data/pixelfed/bootstrap/cache/*
    gosu www-data php artisan migrate --force
fi
```

Right?

I really can't say if this is a good idea or not. Maybe someone else here can join in?

Strubbl:
I am also not sure. Let's leave it out in case of any doubts.

Murazaki:
I believe calling `migrate` would migrate only if necessary though ?
`php artisan migrate --pretend` would also give the content of the possible migration.

In the case you still need to check, something like this might work :

``` bash
has_migrated()
{
  ## convoluted regex that checks for migration name between spaces
  ## followed by dots and a number in brackets, returns No if this exact word follows
  pattern="^\s*[a-zA-Z0-9_]+\s*\.+\s*\[[0-9]+\]\s*(No)$"
  while read -r line; do
    if [[ $line =~ $pattern ]]; then
      echo "${BASH_REMATCH[1]}"
      return
    fi
  done
}

php artisan migrate:status | has_migrated
```

Strubbl:
> I believe calling `migrate` would migrate only if necessary though ?

Yes, i think so. What happens when the pixelfed app and worker containers are started in parallel and both start a migration?

harmtemolder:
> What happens when the pixelfed app and worker containers are started in parallel and both start a migration?

`worker.apache.sh` sleeps for 5 seconds when a migration is needed, it doesn't actually migrate itself. So no two migrations would start at the same time. Or do you mean, what would happen to the worker if it starts while `start.apache.sh` is still running a migration?

I do agree that the `No` is based on an assumption, so I do like the idea of a regex. Something along the lines of:

```shell
gosu www-data php artisan migrate:status | grep -P '^\s+\w+\s+\.+\s+\[\d+\]\s+No\s+$'
```

contrib/docker/start.apache.sh
```diff
@@ -1,15 +1,39 @@
...
+cd /var/www || exit
+
 # Create the storage tree if needed and fix permissions
 cp -r storage.skel/* storage/
 chown -R www-data:www-data storage/ bootstrap/
 
+if [[ ! -e storage/.docker.init ]]; then
+    echo "Fresh installation, initializing database..."
+    chown www-data:www-data .env &&
+        gosu www-data php artisan key:generate &&
```
intentionally-left-nil, line 13: You should add import:cities here as well

contrib/docker/start.apache.sh
```diff
@@ -1,15 +1,39 @@
...
+gosu www-data php artisan horizon:publish
+gosu www-data php artisan config:cache
+# gosu www-data php artisan cache:clear
+gosu www-data php artisan route:cache
+gosu www-data php artisan view:cache
+
+# Check for migrations
+gosu www-data php artisan migrate:status | grep No && migrations=yes || migrations=no
+
+if [ "$migrations" = "yes" ]; then
```
intentionally-left-nil, line 30: This doesn't give control to the admin to decide when to run the db migrations - which ... I see two sides of the argument. Maybe add an environment variable? e.g. ALLOW_AUTO_MIGRATION or something and only run the code here if that's set? That would give admins the ability to manually run the migration (e.g. after they've backed up the database)

intentionally-left-nil (commented):
I did something similar for my setup (although I just ran it as a one-off container rather than every boot:)
```
  initialize:
    image: pixelfed
    profiles:
      - setup
    env_file:
      - ./secrets/pixelfed/dev.env
      - ./secrets/pixelfed/init.env
    volumes:
      - './secrets/pixelfed/dev.env:/var/www/.env'
    networks:
      - db
      - cache
    depends_on:
      db:
        condition: service_healthy
      redis:
        condition: service_healthy
    command:
      [
        '/bin/sh',
        '-c',
        'php artisan migrate && php artisan import:cities && php artisan instance:actor && php artisan config:cache && php artisan route:cache && php artisan view:cache && php artisan passport:client --personal --name=pixelfed -n && (php artisan user:create --name="$$ADMIN_DISPLAY_NAME" --email="$$ADMIN_EMAIL" --username="$$ADMIN_USER_NAME" --password="$$ADMIN_PASSWORD" --is_admin=1 --confirm_email=1 --no-interaction || true)',
      ]
```

so adding something built-in would be nice

intentionally-left-nil:
Oh, and can you change the start.fpm.sh with similar functionality so they stay in sync? (or maybe we should refactor out the common code into a shared script)

dansup:
Superseded by #4844

PR #4185, opened 2023-02-15
Improve Dockerfile

harmtemolder:
Splitting up #4074. This PR only includes the improvements to the Dockerfile:

- `Dockerfile.apache` is split in a couple more `RUN`s so that rebuilding the image locally can reuse cached steps for each
- `apt-get` and PHP clean up are done in their respective steps, instead of at the end
- `php artisan horizon:publish` is removed because this already happens in `contrib/docker/start.apache.sh`
- `.dockerignore` now ignores more unneeded files to keep the image compact

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,15 @@
-data
-Dockerfile
-contrib/docker/Dockerfile.*
-docker-compose*.yml
+# Ignore all markdown files
+*.md
+
+# And these files and folders
+.circleci
+.ddev
 .dockerignore
-.git
-.gitignore
-.env
+.editorconfig
+.env*
+.git*
+.node-version
+contrib/docker/Dockerfile.*
+data
+docker-compose*.yml
+LICENSE
--- a/contrib/docker/Dockerfile.apache
+++ b/contrib/docker/Dockerfile.apache
@@ -4,96 +4,98 @@ ENV COMPOSER_MEMORY_LIMIT=-1
 ARG DEBIAN_FRONTEND=noninteractive
 WORKDIR /var/www/
 
-# Get Composer binary
+# Get Composer binary from https://hub.docker.com/_/composer/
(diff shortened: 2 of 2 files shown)


Strubbl (approved):
I really like the ordering and sorting here. 

I would always go for only one RUN command in order reduce the layers to save disk space. Did you compare how many more space is consumed by splitting into multiple RUNs? At the moment without any of your current PRs my image is already at ~1.5GB. 🙈

harmtemolder:
I'm now at ~1.65 GB

One of the reasons I prefer to split these is that I'm working from a metered connection, and every edit later on in the `Dockerfile` would result in me having to download everything from the start again. But I'm open to discussing this

Strubbl:
I also like more layers during development of a dockerfile. When i am done with the dockerfile, in the last step i usually summarize all RUN commands into one command. 
Do you expect any further changes to the Dockerfile after this is merged? Or is this already meant w.r.t. to a distribution later on via docker hub?

A concern of mine would be that the first layer gets outdated and thus bears a security risk, because in the first RUN we have the apt update & upgrade commands. I would prefer to build this layer more often in order to get all package updates.

intentionally-left-nil:
If you are going to make changes to the dockerfile, can you make the equivalent changes to the FPM dockerfile? They should stay in sync as much as possible

Strubbl:
Why are we not removing the FPM Dockerfile and keep only one Dockerfile?

intentionally-left-nil:
Given that both apache and nginx/fpm are 1st-class supported platforms, I think it's important to keep both. Otherwise you'll end up with one or the other decaying over time.

Not only does it provide a reference point for any non-docker installs, but it also can directly be used. E.g. my personal docker-compose.yml directly takes the Dockerfile.fpm in this repository to generate the fpm container

dansup:
Superseded by #4844

PR #4186, opened 2023-02-15
Improve docker-compose.yml

harmtemolder:
Splitting up #4074. This PR only includes the improvements to `docker-compose.yml`:

- The database service now uses the `.env` file's `DB_*` parameters, so no implementation-specific `MYSQL_*` parameters are needed
- Added `depends_on` condition to make sure MariaDB is fully up before starting Pixelfed, based on the respective `healthcheck` property of the database service
- Added names to all services and networks for easier identification
- Also updated `.editorconfig` to match the code style I encountered in all of the files I opened


docker-compose.yml
```diff
@@ -1,5 +1,4 @@
----
-version: '3'
+version: '3.7'
```
Strubbl, line 1: is there any specific reason to go for 3.7 instead of 3? I only fear that we have to maintain that number in future if we are more specific with 3.7 here instead of a generic 3.

.editorconfig
```diff
@@ -1,9 +1,17 @@
...
 
 [*]
-indent_size = 4
-indent_style = tab
-end_of_line = lf
 charset = utf-8
-trim_trailing_whitespace = true
+end_of_line = lf
+indent_size = 4
+indent_style = space
```
Strubbl, line 7: I am not sure if this change is going to be accepted. This file is not directly coupled to the docker process, or?

Strubbl (approved):
Thanks a lot for your changes, especially all the comments in the docker-compose file are very helpful.

harmtemolder:
`version: '3'` actually means `version: '3.0'`, not the latest `3.x`: https://docs.docker.com/compose/compose-file/compose-versioning/#v2-and-v3-declaration. That is why I specify the minor version

Regarding `.editorconfig`: the annoying thing was that different standards were used in different file types, so whenever I auto-formatted things I was messing up lines that didn't need messing up. This fixes that. But if you think this is major enough to warrant its own PR, I'll split it up again. Because you're right, it's not actually Docker functionality.

Strubbl:
I am fine with this change. Hopefully the tabs/spaces change does not block the merging of the docker-related changes. But let's wait, what the maintainer says.

dansup:
Superseded by #4844

PR #4187, opened 2023-02-15
Clean up env files

harmtemolder:
Splitting up #4074. This PR only includes the improvements to the `.env` files:

- The `.env.example` file now contains all possible variables set to their respective defaults, but commented. This means that as long a user can copy this file to `.env` and uncomment the variables that they would like to change from their defaults.
- `.env.docker` is a copy of `.env.example` with just the changes that will enable a basic Docker setup
- I also figured a link to the docs (https://docs.pixelfed.org/technical-documentation/config/) would be helpful


Murazaki:
First of all, thanks, this is a huge amount of work to uncover all of those hidden variables.

The dissociation of `.env.docker` and `.env.example` is pretty important here, as it might be complicated to find your information in all of this.

I think there's a little more to do to help people find what they need to change here:

- I think the `.env.docker` might be better off with the least possible amount of clutter. Anything not linked to first run of the instance or supplementary default services (like redis, mysql, s3, mail) should not stay there, but the rest might as well be added from the `.env.example`. Users can still add supplementary `.env` files to the `docker-compose.yml` if necessary
- `.env.example` will definitely need more documentation than that. A non informed user might think they have to fill it if not documented properly. This is a huge work. We also need to take into account that non docker user might be using this file too.
- With the amount of parameters, I'm compelled to propose a possible split of those into different .env files that we could store in an `env` folder. This would help a lot in readability and functionality splitting, but requires decisions from the core team.

dansup:
Superseded by #4844

PR #4188, opened 2023-02-15
Update backup.php

Strubbl:
do not include cache directories into the backup

### DIFF
--- a/config/backup.php
+++ b/config/backup.php
@@ -28,6 +28,7 @@ return [
                  */
                 'exclude' => [
                     base_path('.git'),
+                    base_path('cache'),
                     base_path('vendor'),
                     base_path('node_modules'),
                 ],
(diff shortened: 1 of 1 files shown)


mitexleo (dismissed):
It makes the backup files size smaller. Thanks for the PR !

Strubbl:
I just tested it and it did not work as expected at first. Therefore i adjusted the path to the correct absolute path of the public cache directory.

On my single user instance it reduced the backup file size from 1.5 GB to ~100MB, because the avatars are not backed up anymore:

```
$ ls -lh
…
-rw-r--r-- 1 root root 1.5G Mar  4 22:40 2023-03-04-21-38-35.zip
-rw-r--r-- 1 root root  82M Mar  4 22:51 2023-03-04-21-49-59.zip
```

The backup 2023-03-04-21-38-35.zip is withouth this patch and 2023-03-04-21-49-59.zip was created with this patch.

dansup (approved):

PR #4237, opened 2023-03-17
Portfolio Federation + Customization

dansup:
### Portfolio Improvements
- ActivityPub Support
- Optional Portfolio RSS Feed
- Light/Dark mode
- Customize Background/Text color
- Accessibility Improvements

### Admin Report Emails
Enable admin report emails to get notified when new autospam detections or spam reports are created.

Portfolio improvements are funded by our [NGI Grant](https://nlnet.nl/project/Pixelfed-Groups/)


app/Http/Controllers/PortfolioController.php
```diff
@@ -308,11 +452,117 @@ public function storeCurated(Request $request)
...
+        }
+
+        $portfolio = Portfolio::whereUserId($user->id)->where('active', 1)->firstOrFail();
+        $metadata = $portfolio->metadata;
+        $baseUrl = config('app.url');
+        $page = $request->input('page');
+
+        $res = [
+        	'@context' => 'https://www.w3.org/ns/activitystreams',
+        	'id' => $portfolio->permalink('.json'),
```
trwnh, line 539: why `.json`? this probably shouldn't be in there

app/Http/Controllers/PortfolioController.php
```diff
@@ -308,11 +452,117 @@ public function storeCurated(Request $request)
...
+        	'type' => 'OrderedCollection',
+        	'totalItems' => isset($metadata['posts']) ? count($metadata['posts']) : 0,
+        ];
+
+        if($request->has('page')) {
+        	$start = $page == 1 ? 0 : ($page * 10 - 10);
+        	$res['id'] = $portfolio->permalink('.json?page=' . $page);
```
trwnh, line 546: again, why `.json` in id?

app/Http/Controllers/PortfolioController.php
```diff
@@ -308,11 +452,117 @@ public function storeCurated(Request $request)
...
+        	$res['type'] = 'OrderedCollectionPage';
+        	$res['next'] = $portfolio->permalink('.json?page=' . $page + 1);
```
trwnh, line 548: again, why `.json` in id?

app/Http/Controllers/PortfolioController.php
dansup: It's intentional, and differentiates from the `.rss` feed.

app/Http/Controllers/PortfolioController.php
```diff
@@ -308,11 +452,117 @@ public function storeCurated(Request $request)
...
+        	$res['partOf'] = $portfolio->permalink('.json');
```
trwnh, line 549: again, why `.json` in id?

app/Http/Controllers/PortfolioController.php
```diff
@@ -308,11 +452,117 @@ public function storeCurated(Request $request)
...
+        	$res['orderedItems'] = collect($metadata['posts'])->slice($start)->take(10)->map(function($p) {
+        		return StatusService::get($p);
+        	})
+        	->filter()
+        	->map(function($p) {
+        		return $p['url'];
```
trwnh, line 555: if i'm reading this right, this might break if `url` is not present. maybe fallback to `id` in that case?

trwnh: actually wait, isn't this supposed to be rendering the `orderedItems`? in that case, it should use the `id`

app/Http/Controllers/PortfolioController.php
trwnh: we should be using content negotiation via the `Accept` header

PR #4324, opened 2023-04-24
[Fix] Don't count inactive accounts for registration blocking

eitch:
This resolves the issue, where spam accounts which were deleted, block creating new accounts on instances with low max user counts.


eitch:
Closing PR: Wrong source branch

PR #4328, opened 2023-04-24
Add Federated Reports

dansup:


app/Util/ActivityPub/Inbox.php
```diff
@@ -1140,4 +1146,80 @@ public function handleStoryReplyActivity()
...
 		return;
 	}
+
+	public function handleFlagActivity()
+	{
+		if(!isset(
+			$this->payload['id'],
+			$this->payload['type'],
+			$this->payload['actor'],
+			$this->payload['content'],
```
trwnh: isn't content optional? there might not be a comment/reason included

app/Util/ActivityPub/Inbox.php
```diff
@@ -1140,4 +1146,80 @@ public function handleStoryReplyActivity()
...
+		$id = $this->payload['id'];
+		$actor = $this->payload['actor'];
+		$content = Purify::clean($this->payload['content']);
+		$object = $this->payload['object'];
+
+		if(Helpers::validateLocalUrl($id) || parse_url($id, PHP_URL_HOST) !== parse_url($actor, PHP_URL_HOST)) {
+			return;
+		}
+
+		if(!is_array($object) || empty($object)) {
```
trwnh: `!is_array` seems like an unnecessary check, if i understand this correctly. it might be an iri reference if there's only one reported object.

app/Util/ActivityPub/Inbox.php
```diff
@@ -1140,4 +1146,80 @@ public function handleStoryReplyActivity()
...
+			return;
+		}
+
+		if(count($object) > 40) {
```
trwnh: 40 seems quite arbitrary

app/Util/ActivityPub/Inbox.php
```diff
@@ -1140,4 +1146,80 @@ public function handleStoryReplyActivity()
...
+
+		$objects = collect([]);
+		$accountId = null;
+
+		foreach($object as $objectUrl) {
+			if(!Helpers::validateLocalUrl($objectUrl)) {
+				return;
+			}
+
+			if(str_contains($objectUrl, '/users/')) {
```
trwnh, line 1187: this seems like a fragile check that will have to be refactored later

app/Util/ActivityPub/Inbox.php
```diff
@@ -1140,4 +1146,80 @@ public function handleStoryReplyActivity()
...
+				$username = last(explode('/', $objectUrl));
+				$profileId = Profile::whereUsername($username)->first();
+				if($profileId) {
+					$accountId = $profileId->id;
+				}
+			} else if(str_contains($objectUrl, '/p/')) {
```
trwnh, line 1193: another fragile check

PR #4339, opened 2023-05-02
add OpenWebAuth client support

abanink:
[OpenWebAuth](https://zotlabs.org/page/zot/specs+openwebauth) provides a light-weight form of cross-domain authentication between websites on the open web. 
This PR shows how Pixelfed users could log in as visitors on OpenWebAuth enabled services like Hubzilla and the project nicknamed "Streams" or any Fediverse server which implements the server-side of OpenWebAuth. Pixelfed users do not need an account on the other server. 
In practice, it depends on the services and the permissions that the Pixelfed user has on the other server's services; e.g. a user on Hubzilla or Streams could allow a Pixelfed user to post on his wall, upload a picture to his cloud etc.
Note: this is only the 'client' side of the protocol i.e. Hubzilla/Streams users could not log in to Pixelfed.

How this can be tested:

Prerequisites:
- Hubzilla or Streams server ("OWA-S")
- user on Pixelfed

Procedure:
- Log in on Pixelfed
- In the same browser in a new tab: open the login screen of the OWA-S, perform a 'remote authentication' and enter the Pixelfed handle as identifier

Expected outcome:
- The Pixelfed user should now be logged in to the OWA-S as a visitor.

Note: this is just a test scenario for the protocol. 

In a more real-life scenario, resources on Hubzilla/Streams would be available as 
`https://OWA-S/<resource>?zid=<pixelfed handle> 
`
in order for the user having this Pixelfed handle to access the resource on the OWA-S.

The protocol code was borrowed from the Hubzilla code and adapted by me to work with the Pixelfed code. I did not make an effort to re-use existing code in Pixelfed.

### DIFF
--- a/app/Http/Controllers/Magic.php
+++ b/app/Http/Controllers/Magic.php
@@ -1,0 +1,343 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Auth;
+use Illuminate\Http\Request;
+use GuzzleHttp\Client;
+
+class Magic extends Controller
+{
+
+    /**
+     * @brief Generates a random string.
+     *
+     * @param number $size
+     *
+     * @return string
+     */
+    private function random_string($size = 64) {
+        // generate a bit of entropy and run it through the whirlpool
+        $s = hash('whirlpool', (string) rand() . uniqid(rand(),true) . (string) rand(), false);
+
+        return(substr($s, 0, $size));
+    }
+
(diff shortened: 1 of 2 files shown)


dansup:
This is a bit dated, I am working on an OIDC package that will abstract this into its own repo for easier maintenance, and we can do the same for this. I will ping you when it's ready!

abanink:
@dansup great, let me know if I could be of any help (testing, test account etc)

PR #4342, opened 2023-05-03
Staging

dansup:


xundeenergie:
yes! Thank you.

grafik

but... after choosing "#Agontiz" via cursor-keys and pressing \<Enter\>

grafik
and nothing happens.

Klicking on "#Agonitz" opens the search-result-page.

PR #4466, opened 2023-06-12
Import from Instagram

dansup:
With the Import from Instagram feature, you can seamlessly transfer your photos, captions, and even hashtags from your Instagram account to Pixelfed, ensuring a smooth transition without losing your cherished memories or creative expressions.

### For Users

Follow the [Import Help Center](https://pixelfed.social/site/kb/import) instructions, limits and permissions may vary between instances, and some instances may opt to not enable this feature.

### For Admins

You don't need to do anything, this feature is enabled by default!

---

This feature was funded by an [NLnet](https://nlnet.nl/project/Pixelfed-Groups/) grant.


x64x2 (approved):

app/Http/Controllers/ImportPostController.php
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    public function getImportedPosts(Request $request)
+    {
+        abort_unless(config('import.instagram.enabled'), 404);
+
+        return ImportStatus::collection(
+            ImportPost::whereProfileId($request->user()->profile_id)
+                ->whereNotNull('status_id')
+                ->cursorPaginate(9)
```
ThisIsMissEm, line 82: Should this be 9, or should it be 10, which is commonly used in other places throughout the codebase? (I'm guessing 9 because the UI is showing a 3x3 grid?)

app/Http/Controllers/ImportPostController.php
```diff
@@ -0,0 +1,298 @@
...
+                ];
+            })->toArray();
+            $ip->creation_date = $c->count() > 1 ? now()->parse($file['creation_timestamp']) : now()->parse($media[0]['creation_timestamp']);
+            $ip->creation_year = now()->parse($ip->creation_date)->format('y');
+            $ip->creation_month = now()->parse($ip->creation_date)->format('m');
+            $ip->creation_day = now()->parse($ip->creation_date)->format('d');
+            $ip->save();
+
+            ImportService::getImportedFiles($pid, true);
+            ImportService::getPostCount($pid, true);
```
ThisIsMissEm, line 150: Why run these on each iteration? Wouldn't it only be necessary once all posts have been imported?

app/Http/Controllers/ImportPostController.php
dansup: Yes, it's because of the 9x9 grid!

app/Http/Controllers/ImportPostController.php
dansup: To prevent race conditions if multiple windows are open and processing imports. These aren't too expensive to refresh, so I figured better off safe than sorry!

resources/views/site/help.blade.php
```diff
@@ -254,6 +254,22 @@
...
+    <div class="col-12 col-md-6 mb-3">
+      <a href="{{route('help.import')}}" class="text-decoration-none">
+        <div class="card">
+          <div class="card-body">
+            <p class="py-1 text-center">
+              <i class="far fa-file-import text-lighter fa-2x"></i>
+            </p>
+            <p class="text-center text-muted font-weight-bold h4 mb-0">Import</p>
+            <div class="text-center pt-3">
+              <p class="small text-dark font-weight-bold mb-0">How to Import from Instagram</p>
```
ThisIsMissEm, line 266:
```suggestion
              <p class="small text-dark font-weight-bold mb-0">How to import from Instagram</p>
```

resources/views/site/help/import.blade.php
```diff
@@ -0,0 +1,94 @@
...
+            <span>Follow the Instagram instructions on <strong>Downloading a copy of your data on Instagram</strong> on <a href="https://help.instagram.com/181231772500920" class="font-weight-bold">this page</a>. <strong class="text-danger small font-weight-bold">Make sure you select the JSON format</strong></span>
+        </li>
+        <li class="mb-2">
+            <span>Wait for the email from Instagram with your download link</span>
+        </li>
+        <li class="mb-2">
+            <span>Download your .zip export from Instagram</span>
+        </li>
+        <li class="mb-2">
+            <span>Navigate to the <a href="/settings/import" class="font-weight-bold">Import</a> settings page</span>
```
ThisIsMissEm, line 25: I'd suggest having the link cover both words

```suggestion
            <span>Navigate to the <a href="/settings/import" class="font-weight-bold">Import settings</a> page</span>
```

resources/views/site/help/import.blade.php
```diff
@@ -0,0 +1,94 @@
...
+        </li>
+        <li class="">
```
ThisIsMissEm, line 27:
```suggestion
        <li>
```

We can remove the empty class attribute here

resources/views/site/help/import.blade.php
```diff
@@ -0,0 +1,94 @@
...
+            <div class="font-weight-bold">{{ config('import.instagram.allow_video_posts') ? '✅' : '❌' }}</div>
+        </div>
+    </div>
+
+    <hr class="mb-4" />
+
+    <p class="text-center font-weight-bold mb-0">Import Permissions</p>
+    <p class="text-center small">Who is allowed to use the Import feature</p>
+
+    <div class="list-group">
```
ThisIsMissEm, line 64: Is it possible to have no limits at all? e.g., all accounts can perform imports? If so, it might be worth looking at copy that handles that "Anyone can use Import on this server", rather than listing the individual potential rules.

resources/views/site/help/partial/sidebar.blade.php
```diff
@@ -21,12 +21,18 @@
       {{-- <li class="nav-item {{request()->is('*/direct-messages')?'active':''}}">
         <a class="nav-link font-weight-light text-muted" href="{{route('help.dm')}}">{{__('helpcenter.directMessages')}}</a>
       </li> --}}
+      {{-- <li class="nav-item {{request()->is('*/tagging-people')?'active':''}}">
+        <a class="nav-link font-weight-light text-muted" href="{{route('help.tagging-people')}}">{{__('helpcenter.taggingPeople')}}</a>
+      </li> --}}
```
ThisIsMissEm, line 26: This change seems unrelated?

ThisIsMissEm (commented):
Looks good, just a few small comments. One thing that might be worth looking at is a script to clean up failed imports (e.g., the file was uploaded, but the import was not finalised)

app/Http/Controllers/ImportPostController.php
ThisIsMissEm: Right, but these would only kick-in on the start of the import, wouldn't they?

PR #4470, opened 2023-06-12
Add Redis prefix

lao9s:
When you have multiple Laravel projects on the server and a Redis server, running `php artisan horizon` will execute jobs from other Laravel projects that also have queue/horizon.

To avoid this, we need to set a Redis prefix in `config/database.php`, which is what I did.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Database\DBAL\TimestampType;
+use Illuminate\Support\Str;
 
 return [
     /*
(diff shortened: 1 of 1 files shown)


HermanRonk (commented):
Wouldn't the "HORIZON_PREFIX" from the .env configuration be enough to prevent the issue you are trying to solve? (https://docs.pixelfed.org/technical-documentation/config/#horizon_prefix)

dansup:
@lao9s As @HermanRonk mentioned, the `HORIZON_PREFIX` can solve this, as well as you can change `config/database.php` as you please, I don't see the point of polluting the already complex configuration files.

Issue #2935, opened 2021-09-15
HTTP Signature failing against GoToSocial and Mastodon with secure mode set to ON 


tsmethurst:
I did a bit of investigation yesterday and today about the following issue:

https://github.com/superseriousbusiness/gotosocial/issues/131

The tl;dr is that HTTP Signatures don't seem to be working for Pixelfed => secure instances, at least for the instance actor fetching.

I logged a bunch of my findings in the linked issue, so I hope that helps to debug :)

tsmethurst:
For reference: https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures#section-2.3

See particular 4.1 under that heading, to do with concatenating values from same-key headers.

I *think* Pixelfed is not comma-separating values of headers with the same key, but i'm not entirely sure

tsmethurst:
As far as I can tell this is still not fixed unfortunately, see my comment here: https://github.com/pixelfed/pixelfed/commit/a900de2121762cab296c9a6a160492edf71156c4#r56770957

egon0:
Sadly, it's not fixed and prevents a larger adoption of pixelfed in my family and Peergroup 😐

mrvdb:
Is someone still looking at this?  I think not being able to subscribe from gotosocial account to a pixelfed account is a manifestation of this issue?

tsmethurst:
If pixelfed doesn't sign outgoing GET requests, it could be related to that.

I did a bit of a fix for this for gnu social here: https://codeberg.org/GNUsocial/gnu-social/commit/2df30e298727a5097b6712bf053e9402d4fea328

@dansup you might be interested in porting that code into pixelfed, since iirc you used the same gist for your http signature implementation right? :)

anantshri:
@dansup is there something that can be done to help with prioritisation of this issue.

dansup:
@tsmethurst Fixed in #4504 and tested on my staging instance, I can now successfully search and view GtS posts!

tsmethurst:
Nice work :)

Issue #2713, opened 2021-04-04
Cannot fetch profiles from remote Mastodon instances with AUTHORIZED_FETCH: `Undefined index: outbox` 


inga-lovinde:
When I'm trying to search for any beach.city user (e.g. `@anne@beach.city`), I'm getting "An error occured, results could not be loaded. Please try again later." (I tried doing that both on my local instance with v0.10.10, and on pixelfed.de)

In laravel.log, there is a message:
```
production.ERROR: Undefined index: outbox {"userId":4,"exception":[object] (ErrorException(code: 0): Undefined index: outbox at 

[2021-04-04 01:27:20] production.ERROR: Undefined index: outbox {"userId"':4, exception: [object] (ErrorException(code: 0): Undefined index: outbox at /var/www/pixelfed/app/Util/ActivityPub/Helpers.php:470)
[stacktrace]
#0 /var/www/pixelfed/app/Util/ActivityPub/Helpers.php(470): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError (8, 'Undefined index...', '/var/www/pixelf...', 470, Array)
#1 /var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php(29): App\\Util\\ActivityPub\\Helpers::App\\Util\\ActivityPub\\{closure}(Object(Illuminate\\Database\\PostgresConnection))
#2 /var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Database/DatabaseManager.php(371): Illuminate\\Database\\Connection->transaction(Object(Closure))
#3 /var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(261): Illuminate\\Database\\DatabaseManager->__call('transaction', Array)
#4 /var/www/pixelfed/app/util/ActivityPub/Helpers.php(479): Illuminate\\Support\\Facades\\Facade::__callStatic('transaction', Array)
#5 /var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Cache/Repository.php(385): App\\Util\\ActivityPub\\Helpers::App\\Util\\ActivityPub\\{closure}()
#6 /var/www/pixelfed/vendor/laravel/framework/src/Illuminate/Cache/CacheManager.php(392): Illuminate\\Cache\\Repository->remember('ap:profile:by_u...', Object(Illuminate\\Support\\Carbon), Object(Closure))
[...]

PR #4504, opened 2023-06-26
Update ActivityPubFetchService, fix authorized_fetch support

### LINKED ISSUE #1850: ActivityPub GET requests are not signed
ActivityPub GET requests are not signed. This means any Mastodon instance with `AUTHORIZED_FETCH` enabled (which is most of them now that more people are on 3.0.0) cannot follow a Pixelfed user due to the `outbox` property not being sent to unsigned requests.

dansup:
Closes #1850, #2713 and #2935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 - Update AdminApiController, add instance stats endpoint ([89c3710d](https://github.com/pixelfed/pixelfed/commit/89c3710d))
 - Update config, re-add `PF_MAX_USERS` .env variable to limit max users to 1000 by default ([a6d10f03](https://github.com/pixelfed/pixelfed/commit/a6d10f03))
 - Update AdminApiController, fix stats ([5c5541fc](https://github.com/pixelfed/pixelfed/commit/5c5541fc))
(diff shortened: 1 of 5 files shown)


ThatOneCalculator:
As per https://w3c-ccg.github.io/security-vocab/contexts/security-v1.jsonld , Misskey/Calckey actually use `sec:` prefixes for secure fetch signatures: https://codeberg.org/calckey/calckey/src/commit/e53d1f6bdc6fbb8da2d0b75c1b294708e7244058/packages/backend/src/remote/activitypub/misc/contexts.ts#L99-L137

Is there a reason for using `toot:Ed25519Signature`, `toot:Ed25519Key` etc?

trwnh:
`Ed25519Signature` is not the same as `Ed25519Signature2018`. the former is taken from mastodon's context for end-to-end encrypted messaging, which was added in mastodon 3.2.0 (and i'm guessing dansup copied the context directly from mastodon): https://github.com/mastodon/mastodon/blob/9caa0475f891ded573739cd00e7bc915b31abb12/app/helpers/context_helper.rb#L24

additionally, the exact prefixes should not matter; term ids should be expanded to the full IRI as defined by `@context`. if calckey is using `Ed25519Signature2018` for anything, they should be prepared to accept `https://w3id.org/security#Ed25519Signature2018` as the unambiguous id for that.

PR #4545, opened 2023-07-16
Sign-in with Mastodon

dansup:
Allows users to sign-in or sign-up using their Mastodon account.

## Disabled by default

This feature is disabled by default, to enable it you need to add `PF_LOGIN_WITH_MASTODON_ENABLED=true` to your `.env` file, and run `php artisan config:cache && php artisan cache:clear` to cache the change

## Allowed Domains

This feature uses a list of allowed instances (mastodon.social,mastodon.online,mstdn.social,mas.to) by default, to allow other domains, set `PF_LOGIN_WITH_MASTODON_DOMAINS="example.org,example.com,example.net"` in the comma-separated value format in your `.env` file, and run `php artisan config:cache && php artisan cache:clear` to cache the change

## Easy onboarding != email validation

This feature disables email verification by default to improve the onboarding experience, specifically it sets a valid email without requiring a user to wait for the email and click on a button. To disable this, and require email verification, set `PF_LOGIN_WITH_MASTODON_SKIP_EMAIL=false` in your `.env` file, and run `php artisan config:cache && php artisan cache:clear` to cache the change


x64x2 (approved):

app/User.php
```diff
@@ -31,13 +31,7 @@ class User extends Authenticatable
      * @var array
      */
     protected $fillable = [
-        'name',
-        'username',
-        'email',
-        'password',
-        'app_register_ip',
-        'email_verified_at',
-        'last_active_at'
```
eufelipemateus, line 40: Why did you remove this field? Won't it break?

dansup: Yes, this was a typo and is now fixed.

colega:
Hey, I was trying to find any mention about this in [the docs](https://docs.pixelfed.org/running-pixelfed/installation/) but couldn't find anything. Is this considered a stable feature? Can we use it in _production_?

PR #4617, opened 2023-08-27
Correctly handle visiting the website with a fully qualified username

intentionally-left-nil:
Problem:
When you tag someone on the fediverse with their full username - e.g. @myuser@example.com

When the pixelfed website, or mastodon websites or apps tries to visit `example.com/@myuser@example.com`, currently pixelfed returns a 404.
This patch changes the behavior such that both `example.com/@myuser` and `example.com/@myuser@example.com` both return the same content.

## Test plan
visited: domain.com/user, domain.com/@user, domain.com/@user@domain.com
logged in and visited: domain.com/i/web/username/user, domain.com/i/web/username/@user@domain.com

and verified that all the url's load correctly

I also deployed the fix to my pixelfed and validated that clicking a tagged username at least loads the correct webpage!

### DIFF
--- a/app/Http/Controllers/SiteController.php
+++ b/app/Http/Controllers/SiteController.php
@@ -142,6 +142,10 @@ class SiteController extends Controller
 
 	public function legacyWebfingerRedirect(Request $request, $username, $domain)
 	{
+		if($domain == config('pixelfed.domain.app')) {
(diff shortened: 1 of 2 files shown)


dansup:
I'm concerned that this hasn't been tested enough, I will work on refactoring this into the next version and make sure to credit you in the changelog, thanks!

PR #4618, opened 2023-08-27
[Bugfix] Add directory_visibility to filesystems.php public

shleeable:
https://laravel.com/docs/10.x/filesystem

This makes two improvements:
1) Some people are running pixelfed under a dedicated user, with the php-pfm running under www-data.
the `644 to 664` increases the permissions to match `user` RWX to both `user and group`.

2) directory_visibility defaults to private. This change updates the folder permissions to use the public permissions.


intentionally-left-nil:
I'm not sure this does anything? The docs don't mention this setting and also directories should be 755 to begin with:

https://laravel.com/docs/10.x/filesystem#local-files-and-visibility

Instead, can you use the settings from the doc to indicate what you think the desired permissions should be?
[Local Files & Visibility](https://laravel.com/docs/10.x/filesystem#local-files-and-visibility)

> When using the local driver, public [visibility](https://laravel.com/docs/10.x/filesystem#file-visibility) translates to 0755 permissions for directories and 0644 permissions for files. You can modify the permissions mappings in your application's filesystems configuration file:

```'local' => [
    'driver' => 'local',
    'root' => storage_path('app'),
    'permissions' => [
        'file' => [
            'public' => 0644,
            'private' => 0600,
        ],
        'dir' => [
            'public' => 0755,
            'private' => 0700,
        ],
    ],
],
```

intentionally-left-nil:
If the difference is 755 vs 777, that indicates you have a user problem. E.g. something is not writing as www-data

shleeable:
I've updated without permission change. good to merge.

dansup:
As @intentionally-left-nil mentioned, this indicates a problem with your permissions. I'm hesitant to merge this because it doesn't actually fix the underlying issue as proven by pixelfed.social and several other instances. I think we need to better document the job queue and permissions as that is the real fix.

Issue #4518, opened 2023-07-03
Unable to view Reports in admin


justdaj:
I have my first moderation report, but can't access it.  Clicking on the "Open Reports" does nothing.

Console has two errors

Failed to load resource: the server responded with a status of 500 () api/reports/all:1 
Uncaught (in promise) Error: Request failed with status code 500 vendor.js?id=.....

 v0.11.8

![2023-07-03_14-51-00](https://github.com/pixelfed/pixelfed/assets/5741412/c77c0a86-0768-4271-99f4-edb86905d7d5)

 v0.11.8

justdaj:
This is still an issue.  I have another report which I can not view it.  
Now on v0.11.9 of server

Browser reports errors:

![2023-08-25_11-18-53](https://github.com/pixelfed/pixelfed/assets/5741412/3218f984-47c5-4dd1-82aa-b551719a7678)


I've blurred out the IDs just in case that is sensitive info

b2cc:
There's an error in the SQL statement that's used to generate the report overview. In the `laravel.log` the following trace is generated. This is on PostgreSQL 15, offending query is:
```
select * from "reports" where "admin_seen" is null group by "object_id", "object_type" order by "id" desc limit 7
```

Altering the SQL to
```
select * from "reports" where "admin_seen" is null group by "id", "object_id", "object_type" order by "id" desc limit 7;
```
makes the query work.

Logs:
```
[2023-09-09 17:24:37] production.ERROR: SQLSTATE[42803]: Grouping error: 7 ERROR:  column "reports.id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: select * from "reports" where "admin_seen" is null group by ...
               ^ (Connection: pgsql, SQL: select * from "reports" where "admin_seen" is null group by "object_id", "object_type" order by "id" desc limit 7) {"userId":217,"exception":"[object] (Illuminate\\Database\\QueryException(code: 42803): SQLSTATE[42803]: Grouping error: 7 ERROR:  column \"reports.id\" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: select * from \"reports\" where \"admin_seen\" is null group by ...
               ^ (Connection: pgsql, SQL: select * from \"reports\" where \"admin_seen\" is null group by \"object_id\", \"object_type\" order by \"id\" desc limit 7) at /app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:801)
[stacktrace]
[...]

PR #4634, opened 2023-09-03
New Dockerfile, clean up non-working/old Dockerfiles and update of .dockerignore

### LINKED ISSUE #3337: No Public Docker Image?
This app appears to have no public docker image? I don't see `pixelfed` inside docker hub? Do we have to build our own images?

b2cc:
## Introduction
I've been working on a new `Dockerfile` for Pixelfed for a while now to fix https://github.com/pixelfed/pixelfed/issues/3337. I've been tinkering with various methods of creating the `Dockerfile`, splitting it apart in multiple images, trying to run everything rootless etc. but in the end this was just to convoluted to run via standalone `docker`. Maybe I'll transfer what I've learned to a new `docker-compose` setup. For now this all-in-one image should suffice as a starting point for further improvements (I hope).

The container is already built with systems like k8s/Openshift in mind, where access to the container shell or storage might be limited or not available at all. More about this later.

## Implementation
In order to not reinvent the wheel I used `webdevops/php-nginx` as a basis to build on, since it fulfills basically all of the requirements out of the box, and can be customized quite easily. More about the base image can be found here:
 - https://dockerfile.readthedocs.io/en/latest/content/DockerImages/dockerfiles/php-nginx.html
 - https://github.com/webdevops/Dockerfile/blob/master/docker/php-official/8.2/Dockerfile

## Caveats
`webdevops/php-nginx` uses `supervisord` as a processmanager, which sadly means it doesn't follow the usual docker paradigm of one process/one container. Again, for simplicity's sake I went for this route, and it should also work well for local testing without having to spin up multiple containers or mess around with the container networking too much to get the base app running. This means that the container (currently, I'm still investigating workarounds) can't be run rootless.
Also: **DATABASES ARE NOT INCLUDED** - you have to set up the database (MySQL, PostgreSQL,...) and the caching layer (Memcached, Redis,...) separately.

## What's in the box?
`supervisord`, running the following services:
* `nginx` 1.22 / PHP 8.2 / FPM
* `artisan horizon`
* `artisan queue:work`
* `crond` / `syslogd`

### Building the image
At buildtime the `.env` file as well as `/app/storage/app/public` gets linked to `/app/storage`, which will preserve these files/directories during reboots. I also added the custom script `99-init-pixelfed.sh` which will take care of filling the `.env` file, preparing ENVs like `APP_KEY` etc. if they don't exist already. The script also checks for OAuth and ActivityPub settings and runs the necessary `artisan` commands.

### Preparing to run the image
**First, make sure you mount a persistent volume to `/app/storage`!**

Put all needed ENVs for your setup (see: https://docs.pixelfed.org/technical-documentation/config/) into the container environment. The script `99-init-pixelfed.sh` will piece it all together and create the `.env` file. The reason for this is that it enables an easy change of configuration settings in environments like k8s/Openshift where direct file access/manipulation inside the container is often not possible, and any change to an ENV rolls out a new deployment anyways.

The values for `APP_KEY`, `APP_HASHID_SALT` and OAuth keys will only be created once, and persisted in separate files on `/app/storage` inside the image, which makes the application consistent between restarts. (If consistency is not wanted  for some reason the persistent volume can be omitted, which will make the container basically ephemeral).

### Running the image
Errors regarding the database connection(s) or missing tables are expected during the first startup. I have omitted the `artisan` commands to populate the database, since this could be an issue for anyone who wants to use this image in production and relies on controlled database schema migrations.

After the container has started up, run:
```
php artisan migrate --force
php artisan import:cities
```
to initially populate the database.
Now restart the container - from now on everything should work automatically

## Testing / Review
I've run basic test with this image and as far as I can tell all the basics are working. I would be grateful for a review from the Pixelfed developers to catch whatever configuration, PHP modules or packages I might have missed, and would be happy to improve the PR.


b2cc:
I've just discovered an issues with `TRUST_PROXIES` and nginx not sending the correct client IP to the upstream services. However for that I need to tell nginx which IPs to trust, and I thought I'd rely on `TRUST_PROXIES` for that to configure the `real_ip` module.

What format does `TRUSTED_PROXIES` accept? Is it comma or whitespace separated? I couldn't find that in the docs.

EDIT: fixed.

sandervankasteel:
> EDIT: I've just discovered an issues with `TRUST_PROXIES` and nginx not sending the correct client IP to the upstream services. However for that I need to tell nginx which IPs to trust, and I thought I'd rely on `TRUST_PROXIES` for that to configure the `real_ip` module.
> 
> What format does `TRUSTED_PROXIES` accept? Is it comma or whitespace separated? I couldn't find that in the docs.





> 
That's a bit of a gotcha, since that's a Symfony component (and Laravel builds on top of that Symfony component)  But `TRUSTED_PROXIES` is comma separated value and supports the CIDR notation, should you need it :)

Example:
`TRUSTED_PROXIES=127.0.0.1,10.0.0.0/8`

https://symfony.com/doc/current/deployment/proxies.html

b2cc:
@sandervankasteel : Thanks for the answer - then it's already accommodated for in the latest commit I pushed in the morning (I squashed my previous commits so the PR is cleaner). Comma-separated lists work fine now, I tested it with IPv4 and IPv6 addresses. Correct logging of client IPs should also work with complex reverse proxy situations (e.g. multiple proxies in the path) now.

contrib/docker/99-init-pixelfed.sh
```diff
@@ -0,0 +1,68 @@
...
+    | sort > /app/storage/.env
+
+# Create actors if AP is enabled
+ACTIVITY_PUB="${ACTIVITY_PUB:=false}"
+if [[ ${ACTIVITY_PUB} == "true" ]]; then
+  php artisan instance:actor
+fi
+
+# Create OAuth keys if enabled
+OAUTH_ENABLED="${OAUTH_ENABLED:=false}"
```
intentionally-left-nil: This can also be set by PASSPORT_PRIVATE_KEY. See https://github.com/pixelfed/pixelfed/pull/4257

contrib/docker/cron_pixelfed
```diff
@@ -0,0 +1,10 @@
+# .----------- minute (0 - 59)
+# |     .--------- hour (0 - 23)
+# |     |       .------- day of month (1 - 31)
+# |     |       |       .------- month (1 - 12) OR jan,feb,mar,apr ...
+# |     |       |       |       .---- day of week (0 - 6) (Sunday=0 or 7)  OR sun,mon,tue,wed,thu,fri,sat
+# |     |       |       |       |
+# *     *       *       *       *       user           command to be executed
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:
+TERM=xterm
+*/5     *       *       *       *       application    /usr/local/bin/php /app/artisan schedule:run >/dev/null 2>&1
```
intentionally-left-nil: Where do the cron output go? They need to get piped to the docker stdout so errors can be detected and tracked. 

You can use this for inspiration: https://github.com/intentionally-left-nil/pixelfed-docker/blob/main/config/periodic/task.sh

contrib/docker/artisanqueue.conf
```diff
@@ -0,0 +1,18 @@
...
+[program:artisanqueue]
+process_name=%(program_name)s
+command=/usr/local/bin/php /app/artisan queue:work
+user=application
+startsecs = 0
+autostart = true
+autorestart = true
+startsecs = 10
+startretries = 5
+stdout_logfile=/dev/stdout
```
intentionally-left-nil, line 14: Not sure how supervisors works here - if you inspect the docker logs does the worker come through?

b2cc: @intentionally-left-nil : AFAICT this work, e.g. if I interact with the app I get logs like this in the stdout of the container:
```
x.x.x.x - - [05/Sep/2023:08:04:22 +0200] "POST /i/delete HTTP/1.1" 200 43 "https://pixelfed.apps.fedi.at/i/web/timeline/local" "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/117.0"
[php-fpm:access] 127.0.0.1 -  05/Sep/2023:08:04:22 +0200 "POST /index.php" 200 /app/public/index.php 90.266 2048 66.47%
  2023-09-05 06:04:22 App\Jobs\StatusPipeline\StatusDelete ........... RUNNING
  2023-09-05 06:04:22 App\Jobs\StatusPipeline\StatusDelete ..... 220.76ms DONE
  2023-09-05 06:04:22 App\Jobs\MediaPipeline\MediaDeletePipeline ..... RUNNING
  2023-09-05 06:04:22 App\Jobs\MediaPipeline\MediaDeletePipeline  40.82ms DONE
  ```

File: contrib/docker/99-init-pixelfed.sh (hunk 1 above)
Comment (b2cc) [reply contrib/docker/99-init-pixelfed.sh]: Understood, I'll look into how to include this as well.

File: contrib/docker/cron_pixelfed (hunk 2 above)
Comment (b2cc) [reply contrib/docker/cron_pixelfed]: Good catch, I'll update the cronjob to log to stdout as well.

**Comment by b2cc** [comment]
@intentionally-left-nil : I have updated the PR now. 

 - OAuth handling should now be much more robust. I check for files end ENVs, take public and private keys into account, and also have mitigations vs. basic errors in place (e.g. only one of the two keys found, when to generate keys, etc.).

 - added more logging so users see when an error happens

 - crond output redirected to stdout

 - laravel processes already log to stdout, no change

EDIT:
 - discovered some permissions issues on the storage during container restarts, added a fix for that too

 - new issue discovered: currently I get a HTTP/400 on the webfinger resource, e.g. when trying to mention someone on a Mastodon server. The worker delivers the request to the Mastodon server, but the callback from Mastodon to Pixelfed fails (some data redacted):
```
  2023-09-05 20:33:09 App\Jobs\StatusPipeline\StatusLocalUpdateActivityPubDeliverPipeline  RUNNING
  2023-09-05 20:33:09 App\Jobs\StatusPipeline\StatusLocalUpdateActivityPubDeliverPipeline  13.10ms DONE
x.x.x.x. - - [05/Sep/2023:22:34:04 +0200] "GET /.well-known/webfinger?resource=acct:user@pixelfed.domain HTTP/1.1" 400 5 "-" "http.rb/5.1.1 (Mastodon/4.1.6-1; +https://mastodon.server/)"
```
Any ideas what I could be missing here? Additional ENVs or nginx config necessary for webfinger?

contrib/docker/artisanqueue.conf
intentionally-left-nil: thanks for checking!

intentionally-left-nil:
I'm not sure about webfinger. I wonder if we can see any logs from the php side?

b2cc:
@intentionally-left-nil : Any idea how to best debug this? Currently I don't see more than I posted above.
I can see that nginx transfers the request to the php-fpm process, but then I only see the HTTP/400 and that's it. 

Are there any useful debug options for laravel to get behind what's actually causing the error?

x64x2 (approved):

b2cc:
@intentionally-left-nil: Ok I think I worked out all the kinks...

HTTP/400 on webfinger was caused by missing `APP_DOMAIN` in `.env`, federation works now as far as I have tested. I can discover users on other servers and vice versa.

I also reworked the task scheduling again. Wasn't really happy about the cron job. Now it runs as a worker via `supervisord` as well.

I added `->onOneServer()` to `app/Console/Kernel.php` to make scheduling more safe in multi-container environments (task locks are synchronized via scheduler queue), something which can't be accomplished via `crond`. Task logs now also show up in `stdout` together with all the other Laravel logs.
See https://laravel.com/docs/10.x/scheduling#running-tasks-on-one-server for details regarding. `->onOneServer()`

Sqashed all the commits and adjusted the comment again to clean the PR up. From my point of view this is now ready to be merged.

Please let me know if there#s anything else you'd like to see changed.

intentionally-left-nil:
Just FYI you'll need to rebase against pixelfed:staging, since that is Dan's workflow. I'll take a deeper look this weekend!

b2cc:
@intentionally-left-nil : Seems I can't change my source branch in this PR. Should I create a new PR from staging with my changes? Then we can close this.

intentionally-left-nil:
1. git fetch
2. git rebase  upstream/staging (or whatever you have your upstream remote set to)
3. git push -f

So even though your branch name will say b2cc/dev, it will be rebased against staging

b2cc:
Thanks, done. Wasn't aware that actually works, don't use Github that much tbh :sweat_smile:

b2cc:
Small update: I have this image running on a production k8s cluster-backed Pixelfed site since a couple of days now, no problems so far.

b2cc:
@intentionally-left-nil @sandervankasteel : Any news on this? Is there more info necessary or do you require any additional changes or adaptions? I'd appreciate any feedback :pray:

b2cc:
@intentionally-left-nil @dansup : I understand that you're currently pretty busy with development, but some feedback on this PR would be greatly appreciated. I beleive it would be beneficial for the community as well to provide a functional docker image. Thanks for your time and efforts!

EDIT: removed sandervankasteel , was added in error

sandervankasteel:
Hi @b2cc I am afraid I have to disappoint you. I am not a developer or maintainer of Pixelfed. I am a Laravel developer at my day-job and wanted to spin up my own Pixelfed single-user instance. And from that capacity I commented on your `TRUSTED_PROXIES` question.

paulexyz:
Heya to the round and heya @b2cc 👋  I'm happy to see that someone put the time in to build a proper Docker Image for this project. I'd maybe suggest that this image should not also collect and compile the PHP dependencies via Composer - it might also be wise to compile the frontend code as well. I'm aware of the situation that dansup currently is the one who compiles the frontend and then checks that code in the repository before every new release but having that done in an automated way would be very much appreciated. (That might be something for a different PR tough)

Also since you threw Kubernetes in the ring: If one builds fresh container images we might wanna consider scaling in such environments as well. By having a service manager inside a container you kinda build in some obstacles to properly scale in a "cloud" environment. To be constructive here i'm still appreciating a all in one Image (other big projects like gitlab do that too -> omnibus image), but we might wanna make it configurable so the individual components can be scaled individually if wanted.

b2cc:
Hi @paulexyz !
Thanks for your comment/thoughts!
Agreed, there is definitely room for improvement here. Sadly, I don't know much about building the whole frontend code, but it shouldn't be much a problem to add it as a build step. If you can provide an example on how to build it, and which build artifacts need to be included in the final image, then I'm willing to look into it.

Regarding k8s/Openshift: I think we can disregard the idea of recommending this image for such deployments by now. Although it works (I run it on Openshift myself), I agree that it's not an ideal solution.

I only later discovered `nginx-unit`  (https://unit.nginx.org), which would probably be a better approach to run the code (first use some php8-composer image to build all necessary artifacts, then transfer them to `nginx-unit`). Of course a helm chart would be needed, and all the functional components (webserver, horizon, task queues,...) should be provided as separate deployments/services.

For the standalone-approach however I think this image is fine, and should work well for most people.

Anyways, as long as there is no movement or decision on this PR I'd rather not invest more time here, I've already sunk much more efforts into this than originally anticipated. Not sure why though - there seems to be community interest in a working docker build, and there is also another PR regarding this, however even older and probably obsolete by now.

paulexyz:
@b2cc I've currently my own (hacky) build pipeline for building an image. I'd love to help clean up the Container native deployment for this project. I also have an helm deployment which also takes care of deploying all the external dependencies. Wouldn't mind sharing that, but in the current state with no official image I have a hard time doing so...

w4tsn:
Hey there! Thanks for the great work. I forked pixelfed and applied this commit to 0.11.9.

I had to add `WEB_DOCUMENT_ROOT=/app/public"` to the environment to configure php-nginx to start from the correct web-root. I don't know why, yet.

ThisIsMissEm:
So as you've noted, the big problem with the approach used in this PR is the fact that you're running multiple processes within the docker container. This makes it difficult to correctly scale containers.

The generally accepted way of doing this is to run multiple containers: some for web/api traffic, one for the scheduler, and some for the workers. You can see this approach in say, the Mastodon docker-composer setup; Additionally, if using php-fpm, we'd need an nginx container or process running on the host to send traffic to php-fpm.

w4tsn:
> So as you've noted, the big problem with the approach used in this PR is the fact that you're running multiple processes within the docker container. This makes it difficult to correctly scale containers.
> 
> The generally accepted way of doing this is to run multiple containers: some for web/api traffic, one for the scheduler, and some for the workers. You can see this approach in say, the Mastodon docker-composer setup; Additionally, if using php-fpm, we'd need an nginx container or process running on the host to send traffic to puse-cases.If you go this route I'd also add the nginx fully configured for php-fpm to the compose file as you then have this ready to go and if necessary can target a host reverse-proxy to this nginx container specifically configured as entry point for pixelfed.

IMO it's fair to have an all-in-one container (maybe as alternative) for the easy/simple deployment as GitLab does it with it's omnibus container. It's better to have such an off-the-shelf container (for ease of setup) then having no sane  container at all.

If you are really aiming at scalability you'd additionally provide helm charts or swarm config as those usually use completely different config targeted for those use-cases

jippi:
As I was setting up my own pixelfed server, I ended up reworking the entire Dockerfile setup in my PR here: https://github.com/pixelfed/pixelfed/pull/4844

I would love to try and fold some of these improvements into my PR, perhaps as a follow up, if there is interest :)

b2cc:
@jippi @paulexyz : to be honest I consider this PR stale/dead at this point, since there are now three open PRs which tried to get the docker build fixed, and none got merged. There was also no movement on this PR whatsoever in the past couple of months. Not sure why, but I guess project priorities lie elsewhere at the moment. Would have been nice to at least get some word from the dev's about what's going on though.

@ThisIsMissEm : the container is scale-out ready as is - the only process that could be problematic is the scheduler, which I added a fix to make sure the scheduler state is sync'd via redis. See: 
* https://github.com/pixelfed/pixelfed/pull/4634/files#diff-3e91b9df26f6ed49ead694c97ddcddf5f3c3e878390064d0b8849f22d54e5603

jippi:
@b2cc I've chatted with @dansup and offered to take ownership of the Docker related work (since, as you noticed, they tend to sit idle by for months) - I plan to do that and move forward with integrating (in follow up PRs) the improvements and changes from any open Docker related PRs in the repo :)

intentionally-left-nil:
@b2cc I gave up trying to contribute to pixelfed because there's no attempt at building an open source community. It's only whatever Dan's priorities are. I'm sorry your efforts aren't even given the attention they should.

paulexyz:
@intentionally-left-nil @jippi i unfortunately have similar feelings about this project and I’m seriously considering if I can move on hosting this especially since it seems like there is no interest in even remotely merging in some community provided fixes/efforts. Combining that with the lack of fixing existing issues which have a big impact on the project/user base and instead only trying to build new features that IMHO nobody needed or even asked for is very questionable.

I currently don’t have the time&ressources to maintain a (publishable) active fork of this project but having one is kinda a requirement for my own hosting endeavours of this open source project.

PS: I should have looked closer at the repo/issues here before hosting a server - but you’re always smarter after the fact

intentionally-left-nil:
@paulexyz Yeah, I'm also carrying around a set of patches: https://github.com/intentionally-left-nil/pixelfed/tree/main/patches

for my pixelfed fork and feel similarly discouraged about maintaining these patches long-term. Ideally they would all be upstreamed (some version of the fix, not necessarily mine) but there's absolutely no movement.

The real bummer of all of this is that the entire magic of open source is lost this way. Nobody here is trying to commandeer the main direction of pixelfed. Instead, we have encountered bugs, or other pain points and are trying to improve the ecosystem for everyone, for free.

But the cycle only works that way if the maintainers accept patches.

b2cc:
@intentionally-left-nil @paulexyz Yeah was wondering what's going on here, because e.g. issue https://github.com/pixelfed/pixelfed/issues/4518 / PR https://github.com/pixelfed/pixelfed/pull/4642 which I contributed in Sep and which fixes rather important user-facing errors still lingers in `staging`, as well as many other bug fixes. I'd think at least a bugfix release of Pixelfed would be overdue at this point. Anyways, moving on. Thanks to all who commented/helped on this PR!

jippi:
if y'all would want to give my branch a try from https://github.com/pixelfed/pixelfed/pull/4844 I would appreciate it - more than happy to incorporate anything that I might have missed from your branches in there too! :)

b2cc:
@jippi : don't want to get involved here too much anymore, but some recommendations:
 - don't do any `chown` or `chmod` operations at runtime, don't use hardcoded users or UIDs (e.g. `www-data`) in the pod. This will most likely fail in any more involved setup (e.g. rootless,podman,k8s/Openshift). Instead, check for runtime UID and try to accommodate.
 - also: don't `chown -R` [(like here)](https://github.com/pixelfed/pixelfed/pull/4844/files#diff-bfe8bdef67b604f0b1d419ecff17fddc9430f258387c71224f084a0c45a6f6f3) the `storage` folder at startup - people might have huge amounts of data there already. Container startup might take very long in such cases, or fail if startup probes time out and kill the pod. Instead, check permissions / exit with an error.
 - Pixelfed needs some default keys and secrets for federation or it won't work correctly - check for them / generate them / exit if they are missing
 
You can take a look at my startup script, where I already tried to manage some of the above cases:
 - https://github.com/pixelfed/pixelfed/pull/4634/files#diff-ca131308c005d1b17b7f99b32f1e717f9905420c9e3fc7038d90cf9780511ec4
 
All the best!

PR #4643, opened 2023-09-09
Fix potential memory leak due to not calling imagedestroy on GdImage objects

ThisIsMissEm:
I noticed this whilst reading the documentation for the Laravel queues functionality, where it notes that GdImage objects should be explicitly freed to prevent memory leaks in workers: https://laravel.com/docs/10.x/queues#resource-considerations.

Aside: VS Code appears to want to automatically correct formatting on these files, though to minimise diffs & as I'm unfamiliar with project coding style, I thought it best to not modify more than necessary.


ThisIsMissEm:
Note: I've not locally run this code, but it _should_ work, if the PHP documentation is right and I've read the code correctly. I could not see any form of tests for this code.

ThisIsMissEm:
Oh, also, in theory we could maybe move this to using Intervention for image processing in the blurhashes, which iirc, automatically handles this sort of issue. Though that's a much larger change.

dansup (approved):

PR #4650, opened 2023-09-15
Allow session cookies to work on http (insecure)

ThisIsMissEm:
Based on the issue described on discord where even though my instance was running on http://pixelfed.local:8000/, the cookies were being set with the "secure" flag to true, which meant that the cookies wouldn't be available for local development, even if you'd set FORCE_HTTPS_URLS to false.


dansup:
Developers (and admins) can easily edit the `config/*` files, I'm not sure it's worth it to add this as an `.env` variable when it will be rarely used and you can just edit that one line in the config file.

PR #4665, opened 2023-09-25
Add Resilient Media Storage

dansup:
This mode allows you to define up to 2 alternate S3 cloud providers for a total of 3 S3 providers for redundancy.

To enable the primary alternate S3 driver, ensure you set the following .env variables:
- `ALT_PRI_ENABLED=true` 
- `ALT_PRI_AWS_ACCESS_KEY_ID=KEYID` 
- `ALT_PRI_AWS_SECRET_ACCESS_KEY=SECRET`
- `ALT_PRI_AWS_DEFAULT_REGION=REGION`
- `ALT_PRI_AWS_URL=URL`
- `ALT_PRI_AWS_ENDPOINT=ENDPOINT` 

To enable the secondary alternate S3 driver, ensure you set the following .env variables:
- `ALT_SEC_ENABLED=true` 
- `ALT_SEC_AWS_ACCESS_KEY_ID=KEYID` 
- `ALT_SEC_AWS_SECRET_ACCESS_KEY=SECRET`
- `ALT_SEC_AWS_DEFAULT_REGION=REGION`
- `ALT_SEC_AWS_URL=URL`
- `ALT_SEC_AWS_ENDPOINT=ENDPOINT`

### DIFF
--- a/app/Services/ResilientMediaStorageService.php
+++ b/app/Services/ResilientMediaStorageService.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace App\Services;
+
+use Storage;
+use Illuminate\Http\File;
+use Exception;
+use GuzzleHttp\Exception\ClientException;
+use Aws\S3\Exception\S3Exception;
+use GuzzleHttp\Exception\ConnectException;
+use League\Flysystem\UnableToWriteFile;
+
+class ResilientMediaStorageService
+{
+    static $attempts = 0;
+
+    public static function store($storagePath, $path, $name)
+    {
+        return (bool) config_cache('pixelfed.cloud_storage') && (bool) config('media.storage.remote.resilient_mode') ?
+            self::handleResilientStore($storagePath, $path, $name) :
+            self::handleStore($storagePath, $path, $name);
+    }
+
+    public static function handleStore($storagePath, $path, $name)
+    {
+        return retry(3, function() use($storagePath, $path, $name) {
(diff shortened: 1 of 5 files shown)


app/Services/ResilientMediaStorageService.php
```diff
@@ -0,0 +1,66 @@
...
+            $baseDisk = self::$attempts > 1 ? self::getAltDriver() : config('filesystems.cloud');
+            try {
+                $disk = Storage::disk($baseDisk);
+                $file = $disk->putFileAs($storagePath, new File($path), $name, 'public');
+            } catch (S3Exception | ClientException | ConnectException | UnableToWriteFile | Exception $e) {}
+            return $disk->url($file);
+        }, function (int $attempt, Exception $exception) {
+            return $attempt * 200;
+        });
+    }
```
ThisIsMissEm, line 49: This code feels like it's going to randomly scatter files across many different providers, without any way to re-centralise the storage. This should also only be possible to enable if you're also fronting your storage with a custom domain, i.e., ideally your CDN should be doing try_files across each cloud storage provider, and data would be replicated across each provider (at more expense).

I'm doubtful that for most big S3-compatible services that you'd see anything less than 99.9% uptime SLAs (that's a maximum of 87.6 hours of downtime), but most are operating at 99.99% or less than 9 hours of downtime per year) 

(but also, S3 type services aren't designed to ever go down.. https://aws.amazon.com/s3/sla/ )

PR #4677, opened 2023-09-28
Block from menu

viviicat:
This was an issue I've run into a few times. I see someone on the timeline and I want to remove them without visiting their profile. For example if they are a bot posting AI-generated art or whatever. I'd rather not go all the way to their profile to take these actions.

### DIFF
--- a/resources/assets/js/i18n/en.json
+++ b/resources/assets/js/i18n/en.json
@@ -105,6 +105,8 @@
         "viewPost": "View Post",
         "viewProfile": "View Profile",
         "moderationTools": "Moderation Tools",
+        "muteUser": "Mute User",
+        "blockUser": "Block User",
         "report": "Report",
         "archive": "Archive",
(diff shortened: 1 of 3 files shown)


dansup (changes requested):
Can you target the `staging` branch please?

resources/assets/js/i18n/en.json
```diff
@@ -105,6 +105,8 @@
         "viewPost": "View Post",
         "viewProfile": "View Profile",
         "moderationTools": "Moderation Tools",
+        "muteUser": "Mute User",
```
dansup, line 108: You should be adding translations to the `resources/lang/en/profile.php` file so they can be translated to other langs.

The `resources/assets/js/i18n/*.json` files are auto generated by the `resources/lang/en/*.php` files and should never be edited directly.

dansup (changes requested):

PR #4679, opened 2023-09-28
Fix image cropping problem

### LINKED ISSUE #4457: Crop is not working after clicking "Next"
Hi

Firstly, I have to say that I'm new with Pixelfed, i apologize if it is just me not doing well. 

I have an issue with the cropping feature, when I set my cropping as I want, I then click on the "Next" button, and the cropping does not work (I see the image as it was in the first place) 

image

The Next button is loading and then I see the picture, not cropped, without any error messages

image

I'm on Firefox (Ubuntu), up to date (113.0.2), I tried with Chromium as well, not working too. What I do in the cropping section does not save, not even the rotation. I find it weird since it appears to be an "only-me issue". The pixelfed instance I'm on is also up to date (v0.11.8)

Edit : after few tests I was wandering if it happens only with .jpeg, but actually it seems kinda random when I test several images, some work, some not. But the photos from my camera (.jpeg) never crop within pixelfed.

build1024:
Image cropping does not work well when both of the following conditions meet (#4457):
- Object storage is enabled
- MediaStoragePipeline job is executed before images are cropped

This commit delays uploading images to object storage until "Post" is clicked.


dansup:
While I appreciate the work you've done here, we are in the process of refactoring the MediaPipeline to better handle cases like these and also expand to support other formats. I will credit you in the changelog once those improvements are shipped, thanks!

Issue #3596, opened 2022-07-29
Non-explanatory error message when uploading media above 2MB


nehemiagurl:
When trying to upload a picture above 2MB, I get this very non-explanatory error message:
![signal-2022-07-29-173150](https://user-images.githubusercontent.com/109424320/181791184-b089dd77-7c3b-4730-af46-442eeb147d52.jpeg)
(When using an alternative client, e.g. PixelDroid, it gets a generic "500: Internal server error" message, which is also non-explanatory.)

In fact, I only realised the threshold was 2MB through trial and error. The error message alone provides no indication this is a size issue.
Is this limit a result of the Pixelfed code? Of the server settings? Of my user settings? I have no idea - which is part of the problem with such a non-explanatory message.

The best solution to deal with too large images, in my opinion, will be to pop up a prompt asking the user whether they would prefer to remove them or to automatically rescale them so they fit, but as I understand this may be harder to implement and a long-term solution, the error message should be changed to something in this vain:

> **Error: image too large**: The image you uploaded has the size of **xxMB**. Unfortunately, only images up to **yyMB** are supported. Please resize the picture before trying again.

mbliznikova:
Hi, I've created a PR addressing the issue, would really appreciate any feedback/review/comments : https://github.com/pixelfed/pixelfed/pull/4684

PR #4684, opened 2023-10-04
Provide the error message if a file to upload is too large, issue #3596

mbliznikova:
Hello!
This PR refers to the issue [Non-explanatory error message when uploading media above 2MB#3596](https://github.com/pixelfed/pixelfed/issues/3596) .

It handles the case of 413 HTTP status code when uploaded file is too large, adds the explicit error message that the file user is trying to upload is X size, while only files of Y size are supported.

### DIFF
--- a/resources/assets/js/components/ComposeModal.vue
+++ b/resources/assets/js/components/ComposeModal.vue
@@ -1068,6 +1068,16 @@ export default {
 			return App.util.format.timeAgo(ts);
 		},
 
+		formatBytes(bytes, decimals = 2) {
+			if (!+bytes) {
+				return '0 Bytes'
+			}
+			const dec = decimals < 0 ? 0 : decimals
+			const units = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+			const quotient = Math.floor(Math.log(bytes) / Math.log(1024))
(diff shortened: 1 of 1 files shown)


mbliznikova:
Example of the error message that provides the PR when the file size is too large: 

Screen Shot 2023-10-05 at 6 00 40 PM

#3596

aneillans:
Does this also change it for uploads via a client, or do they still get a non-descript 500 error back? (Although clients should be checking the instance info data tbh and controlling that themselves IMO).

mbliznikova:
> Does this also change it for uploads via a client, or do they still get a non-descript 500 error back? (Although clients should be checking the instance info data tbh and controlling that themselves IMO).

Nope, this PR doesn't target uploads via a client

aneillans:
> > Does this also change it for uploads via a client, or do they still get a non-descript 500 error back? (Although clients should be checking the instance info data tbh and controlling that themselves IMO).
> 
> Nope, this PR doesn't target uploads via a client

Did think that was the case; as I say, personally don't think thats a problem as the clients should be checking the node info and doing their own checks - but I do think this improved error is well worth having in the UI.

mbliznikova:
@aneillans I searched for the (mobile) client repo and found this https://github.com/pixelfed/mobile-app , but it says that "The source code will be published after the public release!"

dansup (approved):

PR #4703, opened 2023-10-16
Fix FA-light icons

paulexyz:
When you currently compile the frontend part of the project from scratch all Font Awesome light icons are broken, here's the fix for that.


dansup:
Fixed in the latest commits!

Issue #4708, opened 2023-10-20
Attempt to publish an empty collection through Edit Collections returns 404 response code, not handled in UI


mbliznikova:
_**Pre-requisite**: get the tool to see the requests to API, i.e. Browser Tools - Web Developers tools -> Network tab)_

**Steps to reproduce**
Create and save empty collection, don't publish it yet.
Go to your profile - Collections tab - created empty unpublished collection. 
Click on that collection and then on the "Edit" button - the window with title "Edit Collection" should open.
Check the "Publish" button.

You can see as of now that "Publish" button is enabled when collection has no photos/videos (empty) and has a "Draft" status. 
![Screen Shot 2023-10-20 at 2 45 28 PM](https://github.com/pixelfed/pixelfed/assets/7782090/c4a81bcc-34ba-4280-a9fa-85a319e766c8)

Click on "Publish" button, and when dialog window with question "Are you sure you want to publish this collection?" appears, click on "OK" button there.

**Actual result**
You can see that request to "/api/local/collection/COLLECTION_ID/publish" api returns 404 http response code in the tool that shows the requests to API.
![Screen Shot 2023-10-20 at 2 47 18 PM](https://github.com/pixelfed/pixelfed/assets/7782090/c1705e89-ac82-4da3-b356-01c5818fa16d)

After that the window will close and you will see the "Edit Collection" window again. 
You can click "Save" button and the Edit window in UI will close, but the collection won't be published. 

**Suggestion**
We neither don't have any error/warning messages in UI to let user know that the empty collection wouldn't be published or just that something went wrong, nor we preventing user from publishing empty collection by disabling the Publish button in the **Edit Collection** window. 

When you **create** a collection, the "Publish" button is not available if no photo added to the collection (or if collection has a "Draft" visibility). 
![Screen Shot 2023-10-20 at 2 32 53 PM](https://github.com/pixelfed/pixelfed/assets/7782090/e2490457-3a79-41d7-9f32-e71407869fff)
We can adopt similar approach to the collection editing as well. 
(The current behavior is such that we still can publish non-empty draft collection from Edit Collection tho, but it's visible only to creator).

mbliznikova:
PR has been merged #4709

PR #4709, opened 2023-10-20
Add check if collection is empty in Edit Collection before publishing

mbliznikova:
#4708 
Fix for the issue [Attempt to publish an empty collection through Edit Collections returns 404 response code, not handled in UI ](https://github.com/pixelfed/pixelfed/issues/4708)
Add a check if the collection is not empty in the publishCollection() method as well as setting the condition for "Publish" button to be enabled/disabled depends on check if collection is empty (hope it's not overkill) in CollectionComponent.vue.

As for now I can see that the collections with "Draft" visibility can be published if not empty, but they would be visible only to creators, so I kept such behavior, but please let me know if it's desirable to not publish draft collections at all.

How the fix looks like in my local server ("Publish" button is disabled for an empty collection, enabled for non-empty collection, able to publish the non-empty collection):
Screen Shot 2023-10-20 at 3 03 06 PM

Screen Shot 2023-10-20 at 3 03 40 PM

Screen Shot 2023-10-20 at 3 04 07 PM

Screen Shot 2023-10-20 at 3 04 23 PM

### DIFF
--- a/resources/assets/js/components/CollectionComponent.vue
+++ b/resources/assets/js/components/CollectionComponent.vue
@@ -205,11 +205,19 @@
 
 				<div v-else class="float-right">
 					<button
+					    v-if="posts.length > 0"
 						type="button"
 						class="btn btn-outline-primary btn-sm py-1 font-weight-bold px-3"
 						@click.prevent="publishCollection">
 						Publish
 					</button>
+
+					<button
+						v-else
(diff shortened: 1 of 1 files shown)


essboyer (approved):
Looks good.

dansup (approved):

Issue #4638, opened 2023-09-06
Instagram Import - UTF8 text problem


imto1:
Posts with Persian/Arabic text imported incorrectly and all texts displayed as ASCII chars.
My account is on [pixelfed.social](https://pixelfed.social) instance.

![Screenshot from 2023-09-06 10-20-31](https://github.com/pixelfed/pixelfed/assets/43239919/6ebd546f-c72d-4a8c-a009-c71ccf147433)

unsolaci:
This is probably related to #4479

groovenectar:
Is there any chance you could prepare a minimal example to debug this? E.g., unzip the export, delete everything except the post that doesn't import properly, re-zip and post the zip file for debugging and testing?

imto1:
> Is there any chance you could prepare a minimal example to debug this? E.g., unzip the export, delete everything except the post that doesn't import properly, re-zip and post the zip file for debugging and testing?

OK. I will attach the file when it's ready.

groovenectar:
> OK. I will attach the file when it's ready.

No worries! Someone was already kind enough to make this effort in this thread: https://github.com/pixelfed/pixelfed/issues/4479

Don't do it for now! Sorry for not updating this thread sooner

instalution:
Idea for standard import format! This would allow:

1) PixelFed to focus on PixelFed while import concerns are separated.

2) A standalone program with Composer library so PixelFed can use the imports directly.

3) Some additional recognition for Pixelfed since development could happen under the Pixelfed umbrella, yet the standard format can be used by all Fediverse applications. I'm willing to get started on a repository for this idea if @dansup might eventually agree with the outlined vision and have time to set aside for it.

4) An offset for any potential future quirks with imports that might arise.

For the format, we could use ActivityPub format itself!

Say you have a folder of images:

![image](https://github.com/pixelfed/pixelfed/assets/108841276/0dffa438-bc1b-4cad-994e-0464f5f3f6ca)

[...]

Issue #4479, opened 2023-06-13
Some UTF characters are encoded/decoded wrongly during Instagram export/import


unsolaci:
I tend to use characters like "Σ" instead of "E", 'cause I'm edgy like that, "—" as separators, and sometimes emojis in my Instagram post descriptions. Quite a nasty mix.

Here's an example description from my [Instagram post](https://www.instagram.com/p/CVyPUAgAO88/):
```
I N F V S Σ
👩 @poziomczan_czekoladodylu
—
#light #red #pentagram #smoke #incense #ritual
#witch #polishwitch #witchy #witchyvibes #autumn #fall #halloween
#photoshoot #portrait #portraitphotography #model #femalemodel
```

Here's what it looks like in the exported JSON:
```json
"title": "I N F V S \u00ce\u00a3\n\u00f0\u009f\u0091\u00a9 @poziomczan_czekoladodylu\n\u00e2\u0080\u0094\n#light #red #pentagram #smoke #incense #ritual\n#witch #polishwitch #witchy #witchyvibes #autumn #fall #halloween\n#photoshoot #portrait #portraitphotography #model #femalemodel"
```

And here's the description in the imported [pixelfed.social post](https://pixelfed.social/p/unsolaci/1074896211102011):
```
I N F V S Î£ ð© @poziomczan_czekoladodylu â #light #red #pentagram #smoke #incense #ritual #witch #polishwitch #witchy #witchyvibes #autumn #fall #halloween #photoshoot #portrait #portraitphotography #model #femalemodel
```

This might not be Pixelfed's fault at all, and quite possibly due to the way these characters are encoded by Instagram in its post exports.

I'm not too sure what this encoding is exactly, but when I compare it with HTML entity tables it looks like Pixelfed is converting them correctly.

But why would Instagram convert these characters like that? Again, on IG website and app they all show correctly, so I might be missing something.

realpixelcode:
Perhaps related to issue #1847 (wrongly closed) which also occurs outside the context of Instagram imports.

xavsworld:
It's seems to be a problem from IG's JSON data. I do have the same issue, after importing my own IG stuff. Since I do use French and English on my post descriptions, some chars are simply mess-up. Thats being said, the issue occurs even if you have no emojis.

[...]

PR #4726, opened 2023-11-01
fix: Instagram import broken UTF8 characters

paulexyz:
As discussed in #4479, and #4638 if you import your posts from Instagram you might encounter issues as soon as there are UTF8 characters involved.

This is due to a wrongly encoded JSON which is apparently common for all Meta companies like Facebook and Instagram. A good explanation what's actually going wrong you can find from @ sorashi [here](https://sorashi.github.io/fix-facebook-json-archive-encoding)

To fix this you need to decode and re-encode the text before lettig a JSON parser parse it.

Also slightly unrelated fix is also in here - some media files can be found in the "media/other" folder as well. So i added that to the filter matches

Since it's not wanted to check in compiled/minified JS code i've not included it here. For testing purposes you can check out the [insta-import-utf8](https://github.com/paulexyz/pixelfed/tree/compiled/insta-import-utf8) branch in my repo

### DIFF
--- a/resources/assets/components/AccountImport.vue
+++ b/resources/assets/components/AccountImport.vue
@@ -348,8 +348,16 @@
                 }, 500);
             },
 
+			// Facebook and Instagram are encoding UTF8 characters in a weird way in their json
+			// here is a good explanation what's going wrong https://sorashi.github.io/fix-facebook-json-archive-encoding
+			fixFacebookEncoding(string) {
+				const replaced = string.replace(/\\u00([a-f0-9]{2})/g, (x) => String.fromCharCode(parseInt(x.slice(2), 16)));
+				const buffer = Array.from(replaced, (c) => c.charCodeAt(0));
+				return new TextDecoder().decode(new Uint8Array(buffer));
+			},
+
             filterPostMeta(media) {
-                let json = JSON.parse(media);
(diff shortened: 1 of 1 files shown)


dansup:
@paulexyz Thanks a lot! 

Will be merging this shortly. I'm going to add some logic to separate hashtags without spaces between to fix that edge case too.

dansup (approved):

paulexyz:
@dansup shall we also close the issues related to this?

Walsonde:
> Will be merging this shortly.

@dansup 
I don't mean to rush, I just would like to know, when this fix can be expected. I want my older Instagram posts to appear prior to new posts on Pixelfed. So I can't use Pixelfed before this issue is fixed. I understand that there might be more pressing issues and am quite willing to wait as long as needed.

Thank you for your effort. <3

x64x2 (approved):

Beleggrodion:
I had also the issue (tried it on the pixelfed.de instance) to import all my instagram posts. All newer postings are not imported (only until sept 2021, so 2022+ entries are missing) . I looked into the json and see the utf-8 issue and manually changed the json (replaced the ä ü etc. with ae ue, etc.) but without success.

PR #4729, opened 2023-11-01
fix: CatchUnoptimizedMedia, prevent skipping of imported posts

### LINKED ISSUE #4483: Cannot move imported Instagram-pictures to S3
Hey everyone,

after successfully importing the few Instagram pics I still like today I realized that they aren't moved to my S3-provider (here: Wasabi). Each new pic I upload is nearly instantly stored within S3. When using the "php artisan media:migrate2cloud" to manually push them to S3, the following error appears:

Bildschirmfoto 2023-06-16 um 14 38 08

Pixelfed is installed in /var/www/pixelfed and the directory mentioned is also available

Bildschirmfoto 2023-06-16 um 14 42 45

and linked as everything else is running smoothly in terms of pictures uploaded.

Everything else on there instance works as it should only the imported Instagram-picture are seemingly not uploaded.

The path of a file is, for example, **https://my.pixelfed.tld/storage/m/_v2/555130687540998145/186f813ba-297a89/2saiIi7Tg0Ta/QYb5jKyLg3Ya3B4C0mCZiMGFPHFlKz9bucL5Wdrl.jpg**, seemingly referring to the storage-directory.

Am I missing something or is this maybe just an open issue with the new feature?

Best regards,

Oliver

paulexyz:
When importing posts from Instagram posts that are older as one hour at that time will not be properly processed and uploaded to a remote storage backend because of a where condition in the CatchUnoptimizedMedia Console.

fixes #4483 

For this PR I've removed the constraint. Furthermore was the filter for the field to `skip_optimize`  broken for Postgresql (and maybe others) as well. Hence why I removed it in the select statement at the top and moved it down in the for each loop.


andrewgioia:
Just a general +1 on this PR, would be great to get it merged in. 

I just imported from Instagram and ran into this issue--photos eventually got added locally (after running the artisan command in the associated issue above) but they're stuck at local and not going to S3.

dansup (dismissed):

PR #4739, opened 2023-11-07
Add S3 command to rewrite media urls

dansup:
This is a command to update S3 media urls in the database, useful for when you migrate S3 buckets or want to add a proxy in front of your S3 media.

### Usage

Run `php artisan media:cloud-url-rewrite` and enter the old and new S3 domains when prompted, or you can provide them as arguments like `php artisan media:cloud-url-rewrite old.pixelfed.example.org new.pixelfed.example.org`

### Notes

- Before you can update S3 media urls, you have to configure the new bucket and then re-cache configuration (`php artisan config:cache && php artisan cache:clear` before you run this command
- You need to run `php artisan cache:clear` after the command is run to purge old S3 media urls from your cache


evolucidity:
Late to the party, but thank you @dansup!
Got this working fine on my instance. :-)

x64x2 (approved):

PR #4752, opened 2023-11-11
Experimental home feed

dansup:
This PR provides an experimental home feed api (disabled by default)

The experimental home feed api includes:

- Hashtag follows (posts including a hashtag you follow)
- Improved filtering (de-duped reblogs, show/hide reblogs per account)
- Significant performance improvements (feeds are served from memory, not expensive db queries)

### Config

To enable this feature, set `EXP_CHT=true` in `.env` and run `php artisan config:cache && php artisan cache:clear`

### Notes

- This only affects the `/api/v1/timelines/home` endpoint, the webUI does not consume this endpoint (yet)

----

### Controllers\Api
- [x] ApiV1Controller 

### Jobs\HomeFeedPipeline
- [x] FeedFilterPipeline
- [x] FeedInsertPipeline
- [x] FeedFollowPipeline
- [x] FeedUnfollowPipeline
- [x] FeedRemovePipeline
- [x] FeedWarmCachePipeline
- [x] HashtagInsertFanoutPipeline
- [x] HashtagRemoveFanoutPipeline

### Observers
- [x] StatusHashtagObserver

### Services
- [x] FollowerService
- [x] HashtagFollowService
- [x] HashtagService
- [x] HomeTimelineService

### DIFF
--- a/app/Jobs/StatusPipeline/RemoteStatusDelete.php
+++ b/app/Jobs/StatusPipeline/RemoteStatusDelete.php
@@ -153,7 +153,10 @@ class RemoteStatusDelete implements ShouldQueue, ShouldBeUniqueUntilProcessing
             ->whereObjectId($status->id)
             ->delete();
         StatusArchived::whereStatusId($status->id)->delete();
-        StatusHashtag::whereStatusId($status->id)->delete();
+        $statusHashtags = StatusHashtag::whereStatusId($status->id)->get();
+        foreach($statusHashtags as $stag) {
+        	$stag->delete();
+        }
         StatusView::whereStatusId($status->id)->delete();
         Status::whereInReplyToId($status->id)->update(['in_reply_to_id' => null]);
 
--- a/app/Services/FollowerService.php
+++ b/app/Services/FollowerService.php
@@ -19,6 +19,7 @@ class FollowerService
 	const FOLLOWING_SYNC_KEY = 'pf:services:followers:sync-following:';
 	const FOLLOWING_KEY = 'pf:services:follow:following:id:';
 	const FOLLOWERS_KEY = 'pf:services:follow:followers:id:';
+	const FOLLOWERS_LOCAL_KEY = 'pf:services:follow:local-follower-ids:';
 
 	public static function add($actor, $target, $refresh = true)
 	{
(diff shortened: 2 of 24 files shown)


app/Jobs/StatusPipeline/RemoteStatusDelete.php
```diff
@@ -153,7 +153,10 @@ public function unlinkRemoveMedia($status)
             ->whereObjectId($status->id)
             ->delete();
         StatusArchived::whereStatusId($status->id)->delete();
-        StatusHashtag::whereStatusId($status->id)->delete();
+        $statusHashtags = StatusHashtag::whereStatusId($status->id)->get();
+        foreach($statusHashtags as $stag) {
+        	$stag->delete();
+        }
```
ThisIsMissEm, line 159: Can't this be done by telling laravel to cascade the deletion?

ThisIsMissEm: See: https://laravel.com/docs/10.x/migrations#foreign-key-constraints

app/Services/FollowerService.php
```diff
@@ -212,4 +213,15 @@ public static function delCache($id)
 		Cache::forget(self::FOLLOWERS_SYNC_KEY . $id);
 		Cache::forget(self::FOLLOWING_SYNC_KEY . $id);
 	}
+
+	public static function localFollowerIds($pid, $limit = 0)
+	{
+		$key = self::FOLLOWERS_LOCAL_KEY . $pid;
+		$res = Cache::remember($key, 86400, function() use($pid) {
```
ThisIsMissEm: Caching for a day seems a bit long? Shouldn't this maybe be an hour or few?

app/Jobs/StatusPipeline/RemoteStatusDelete.php
dansup: This isn't about cascading deletes or the database, rather we are doing this to trigger the deleting/deleted events for the StatusHashtagObserver to fanout create/deletes!

app/Services/FollowerService.php
dansup: Good point, I will update this!

expertmanofficial:
It seems to be that this change affected my home feed.

Now instead of posts from people I follow, I see posts from unknown people that I don't follow.

I don't follow any hashtags even.

PR #4794, opened 2023-12-04
Add WebPush

dansup:
Add WebPush and VAPID for Push Notifications


ThisIsMissEm (commented):
When you add this, you can expose the VAPID public key via the instance metadata endpoint (I just did this change in Mastodon 4.3), instead of the `apps/verify_credentials` response. There's also a set of APIs for enrollment.

PR #4798, opened 2023-12-05
Fix: dockerfile libwebp issues

ThisIsMissEm:
See the two individual commits for details; Unfortunately some auto-formatting happened on the dockerfiles so they're a little hard to compare. I've basically removed libwebp7 as a dependency, added wget, enabled the additional database drivers, and changed `apt-get upgrade` to `apt-get dist-upgrade`

Along side this, I took the opportunity to update composer and php to 2.6 and 8.2 respectively. PHP 8.3 isn't compatible with imagick at this time.

### DIFF
--- a/contrib/docker/Dockerfile.apache
+++ b/contrib/docker/Dockerfile.apache
@@ -8,80 +8,78 @@ WORKDIR /var/www/
 COPY --from=composer:2.6 /usr/bin/composer /usr/bin/composer
 
 # Install package dependencies
-RUN apt-get update \
-  && apt-get upgrade -y \
-#  && apt-get install -y --no-install-recommends apt-utils \
-  && apt-get install -y --no-install-recommends \
-## Standard
-      locales \
-      locales-all \
-      git \
-      gosu \
-      zip \
-      unzip \
-      libzip-dev \
-      libcurl4-openssl-dev \
-## Image Optimization
-      optipng \
-      pngquant \
-      jpegoptim \
-      gifsicle \
-## Image Processing
-      libjpeg62-turbo-dev \
-      libpng-dev \
-      libmagickwand-dev \
-# Required for GD
-      libxpm4 \
-      libxpm-dev \
-      libwebp7 \
-      libwebp-dev \
-## Video Processing
-      ffmpeg \
-## Database
-#      libpq-dev \
-#      libsqlite3-dev \
-      mariadb-client \
-# Locales Update
+RUN \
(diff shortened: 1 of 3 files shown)


andre-dierker:
I think you missed to update the fpm config to the new php version socket https://github.com/pixelfed/pixelfed/blob/4bb97e1547202365bbd570ce095a0558c77a4d3b/contrib/nginx.conf#L34

ThisIsMissEm:
> I think you missed to update the fpm config to the new php version socket
> 
> https://github.com/pixelfed/pixelfed/blob/4bb97e1547202365bbd570ce095a0558c77a4d3b/contrib/nginx.conf#L34

So I've tried that, I still can't get the server to respond, just get `curl: (56) Recv failure: Connection reset by peer` both from outside and inside the container when trying to curl :9000, which should be the port it's using.

I've tried with both Apache and NGINX-FPM

ThisIsMissEm:
Also the `/contrib/nginx.conf` file doesn't appear to be used by Docker, nor the `/contrib/docker-nginx.conf` file.

ThisIsMissEm:
Okay, I managed to get the Apache one to work, it uses port 80 not 9000 like nginx-fpm (I didn't notice this before)

x64x2 (approved):

Strubbl:
Just for reference: The imagick with php 8.3 seems to be documented at https://github.com/Imagick/imagick/issues/643

Strubbl (approved):
Thank you for your patches, i tried them with my setup and the image builds without any issues and the container is running fine.

Murazaki:
Hi, the hadolint linter did not like the usage of `dist-upgrade` , but it seems like [the rule is being deleted](https://github.com/hadolint/hadolint/issues/562). I deactivated it from the Dockerfile linting on my side.

Murazaki:
https://github.com/Murazaki/pixelfed/commits/fix/dockerfile-8.2-fpm-with-nginx -> Contains a nginx stack that works, and I made some cleanup after.
I've seen that @ThisIsMissEm you did the same thing as me in unifying the flavors in github actions in your remote, so I figured the lack of difference between fpm and apache can help us unify a bit more. The cleanup is in a different commit.

jippi:
As I was setting up my own pixelfed server, I ended up reworking the entire Dockerfile setup in my PR here: https://github.com/pixelfed/pixelfed/pull/4844

dansup:
Superseded by #4844

PR #4815, opened 2023-12-09
fix: video upload broken

Quadrubo:
This fixes the upload of:
- multiple videos
- multiple photos + videos

The `ComposeController.php` only checks for the possible media types "photo", "video" and "photo:album". The types "video:album" and "photo:video:album" are missing here.

I added these types to allow for the upload of multiple videos in a single post and mixed posts with photos & videos.


Quadrubo:
Nevermind, this breaks all of Pixelfed for me. Can't delete the post and the Frontend is broken.

Issue #4385, opened 2023-05-16
No access to private posts on remote instance?


kr428:
Setup: I'm on Mastodon (or Friendica) and following a pixelfed user with a "private" profile. In this environment, I cannot see most of this contacts posts. Why?

- Looking at this user from my Mastodon (or Friendica) client, I only see the posts that reached my instance, and at some point the system says "go to remote instance to see more".
- Unfortunately, clicking on the profile link, I only see "Account is private" and pixelfed prompting me to log in.

This however, doesn't work: I don't have a pixelfed account on that platform, and even if I had, I wouldn't necessarily be following this particular account. I can't use my Mastodon or Friendica login credentials on pixelfed either. So at the moment, for private profiles like this, "old" data is inaccessible from remote.

This is a difficult situation to be left in. You see any way to handle this...?

abanink:
This is exactly a problem that [OpenWebAuth](https://zotlabs.org/page/zot/specs+openwebauth) solves.

I even made a proof-of-concept implementation of Pixelfed as a OpenWebAuth server (authorizing access to remotely authenticated users).
Now the biggest issue is that Pixelfed doesn't have the concept of authorized users who have not logged in to the Pixelfed server. In my proof-of-concept, to simulate this, I added a constraint that the authorized user must be following the user whose post he wants to see (instead of 'This account is private') when retrieving the posts. But then all links from that page fail because they lack support for remotely authenticated users as well, that's why it's just a proof of concept and not really usable atm.

But it's a way to handle this situation.

trwnh:
some form of cross-domain authorization is needed, yeah. not necessarily openwebauth, though -- you could use indieauth instead, for example. in general, you basically prove to the remote website who you are and then it shows you private stuff that you're allowed to see. but it would indeed require a lot of changes and also require agreement from the rest of the ecosystem on what to actually use to establish your identity across domains.

kr428:
[...]

Issue #4369, opened 2023-05-10
Federation questions


h-2:
I have just "moved" one of my Mastodon accounts to Pixelfed, and have some questions regarding interaction with Mastodon accounts:

1. I don't seem to have reposts in my Homefeed of people that I follow. Is this always the case, or is this just for Mastodon-users I follow? Can I change this behaviour somewhere?
2. Sometimes Pixelfed reports that a post has N comments, but it won't show all of them. See an example below. This is both true for the total number and also for nested replies, where it might say that there are two but then only show one. Fedilab shows the exact same comments, but reports differently (only top-level comments are counted, and there are no reply-counts to replies, so I can't compare).

![image](https://github.com/pixelfed/pixelfed/assets/2996863/06f0e79f-f3ee-405d-bf37-996b7b2ba255)

mitexleo:
> 1. I don't seem to have reposts in my Homefeed of people that I follow. Is this always the case, or is this just for Mastodon-users I follow? Can I change this behaviour somewhere?

It's same for all users not only Mastodon . It's WIP. Will be fixed in this weekend I hope.

SteveDinn:
I am not seeing media-containing posts (not re-posts) from Mastodon users that I follow from my Pixelfed account. Is this related (i.e., a "federation issue") or expected behaviour?

mitexleo:
> I am not seeing media-containing posts (not re-posts) from Mastodon users that I follow from my Pixelfed account. Is this related (i.e., a "federation issue") or expected behaviour?

No ... Looks like an issue with your instance!

SteveDinn:
Yeah, my problem must be something different and weird. I noticed that I only have that problem with *some* accounts, but not all.

Moonstone3010:
For testing, I created several accounts on different Pixelfed instances. I found an account with a large number of image posts on Mastodon and displayed this account sequentially on each Pixelfed instance on which I am logged in. The effect is that I see different content from each account. I've searched for information on how all this federation stuff should work, but got nothing but headaches.
Could someone at least try to explain why I see different posts of the same account from each instance?

mitexleo:
> For testing, I created several accounts on different Pixelfed instances. I found an account with a large number of image posts on Mastodon and displayed this account sequentially on each Pixelfed instance on which I am logged in. The effect is that I see different content from each account. I've searched for information on how all this federation stuff should work, but got nothing but headaches. Could someone at least try to explain why I see different posts of the same account from each instance?

[...]

Issue #3679, opened 2022-10-03
Can't see posts from external instances, global feed is empty


mackiexx:
I can't see post from the global feed. When i search for the user manually, nothing shows on the user's page and followers ,following and post shows "0" even though on the main website, the user has many post and followers. I have the activitypub on

holajsh:
Probably same issue as #3115.
This is imho a major showstopper for the adoption of the product.

TiiFuchs:
I have the same problem. Can't see any posts of external instances. I can search and follow them, and they get a notification, but if they follow back, I get nothing...

There are no errors in the logs or the js console and I can't figure out if there is a configuration I missed, that need fixing...

mitexleo:
I have the same problem  too

ghost:
> I have the same problem. Can't see any posts of external instances. I can search and follow them, and they get a notification, but if they follow back, I get nothing...
> 
> There are no errors in the logs or the js console and I can't figure out if there is a configuration I missed, that need fixing...

Same effect here - searching them is okay, following them, too - but no updates are shown in my timeline, yet my instance just has my own pics in its stream only.

vanlueckn:
I have the same problem

mattlqx:
As well it doesn't scape their post/follow counts on the remote server...

<img width="294" alt="Screenshot 2022-12-17 at 10 48 23 AM" src="https://user-images.githubusercontent.com/739851/208254743-5e564470-2249-4e4d-bc0f-2bbc8ffc2425.png">

You're just left with confirmation that someone exists, and that's it.

eobet:
I’m surprised this is still a problem. Is Pixelfed going the Bluesky route? Claiming to be distributed but then not fully supporting it?

I just followed my first user, and was confused to see that when I pasted their url into the search bar on pixelfed.social it appeared as if they had 0 posts, which I know when viewed from their pixelfed server isn’t true.

How is this not a massive showstopper which should be prioritized before adding any new features? Are we only meant to browse new photos, not older ones?

PR #4825, opened 2023-12-12
Pixelfed as cross-domain Openwebauth server

abanink:
This PR turns Pixelfed into an [OpenWebAuth ](https://zotlabs.org/page/zot/specs+openwebauth) server, meaning that any service with OpenWebAuth client support can log in to a Pixelfed instance as a user that did not previously exist on the Pixelfed instance.
Examples: another Pixelfed server [with OpenWebAuth client support](https://github.com/pixelfed/pixelfed/pull/4339), Hubzilla, (streams), Friendica, Mastodon [with OpenWebAuth client support](https://github.com/mastodon/mastodon/pull/25012), and more will follow.

The immediate value is that it solves issues regarding people on other Pixelfed servers not being able to see older posts since they started following someone, like [here](https://github.com/pixelfed/pixelfed/issues/4369), [here](https://github.com/pixelfed/pixelfed/issues/3679) or [here](https://github.com/pixelfed/pixelfed/issues/4385).
It also respects 'Private' accounts out of the box.
All this is demo-ed in [this video](https://vimeo.com/877251006).

In the longer run, all OpenWebAuth servers/clients could make up for a giant intranet of Fediverse services only requiring a single sign on.

What I am unsure of, is the functionality that could or should be provided, like commenting, following etc. for users logged in through OpenWebAuth. Some functionalities are already disabled in this PR (like editing the profile) to indicate that not everything a local user can do makes sense for a remotely logged in user in the current state of Pixelfed; however it is still possible through the user interface to perform actions that don't make sense to me at this moment so that part is incomplete. I was hoping that this PR could start a discussion that provides insights how to progress on this - although not actually part of the remote login feature.

### DIFF
--- a/app/Http/Controllers/OpenWebAuthController.php
+++ b/app/Http/Controllers/OpenWebAuthController.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use App\Services\BouncerService;
+
+class OpenWebAuthController extends Controller
+{
+    public function showLoginForm() {
+
+		if(config('pixelfed.bouncer.cloud_ips.ban_logins')) {
+			abort_if(BouncerService::checkIp(request()->ip()), 404);
+		}
+
+        return view('auth.openwebauth');
+    }
+
+    public function openWebAuth(Request $request) {
+        $handle = $request->input('handle');
+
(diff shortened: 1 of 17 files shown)


x64x2 (approved):
Well done

dansup:
As I mentioned in #4339, I am working on an OIDC package that will abstract this into its own repo for easier maintenance, and we can do the same for this. I will ping you when it's ready!

PR #4831, opened 2023-12-14
Some UI improvements / fixes in the story's caroussel

Jerome-Herbinet:


x64x2 (commented):
The pink block is okay but the after of the other block is not , the former is much better

Jerome-Herbinet:
> The pink block is okay but the after of the other block is not , the former is much better

@s0mt0chukwu are you thinking that, in the ugol_ki block, the very tiny and left aligned thumb is better displayed ?

dansup:
While I appreciate the sentiment, I don't believe this fits with the current design.

PR #4843, opened 2023-12-27
Add Roles & Parental Controls

dansup:
Roles & Parentals Controls are about 80% ready, however I am pausing work on this to finish some other things. We can safely merge this and it won't interfere with new or existing instances.

~Adds customizable Roles and Parental Controls, this feature is being funded by our NGI Grant~

 ## Funding

This project is funded through the [NGI Assure Fund](https://nlnet.nl/assure), a fund established by [NLnet](https://nlnet.nl) with financial support from the European Commission's [Next Generation Internet](https://ngi.eu) program. Learn more at the [NLnet project page](https://nlnet.nl/project/Pixelfed-Groups).

[NLnet foundation logo](https://nlnet.nl)
[NGI Assure Logo](https://nlnet.nl/assure)

### DIFF
--- a/app/Http/Controllers/Api/ApiV1Controller.php
+++ b/app/Http/Controllers/Api/ApiV1Controller.php
@@ -98,6 +98,7 @@ use App\Jobs\MediaPipeline\MediaSyncLicensePipeline;
 use App\Services\DiscoverService;
 use App\Services\CustomEmojiService;
 use App\Services\MarkerService;
+use App\Services\UserRoleService;
 use App\Models\Conversation;
 use App\Jobs\FollowPipeline\FollowAcceptPipeline;
 use App\Jobs\FollowPipeline\FollowRejectPipeline;
@@ -1623,6 +1624,8 @@ class ApiV1Controller extends Controller
         ]);
 
         $user = $request->user();
+        abort_if($user->has_roles && !UserRoleService::can('can-post', $user->id), 403, 'Invalid permissions for this action');
+
         AccountService::setLastActive($user->id);
 
         if($user->last_active_at == null) {
@@ -1792,6 +1795,7 @@ class ApiV1Controller extends Controller
         abort_if(!$request->user(), 403);
 
         $user = $request->user();
+        abort_if($user->has_roles && !UserRoleService::can('can-post', $user->id), 403, 'Invalid permissions for this action');
         AccountService::setLastActive($user->id);
 
         $media = Media::whereUserId($user->id)
@@ -1831,6 +1835,7 @@ class ApiV1Controller extends Controller
         ]);
 
         $user = $request->user();
+        abort_if($user->has_roles && !UserRoleService::can('can-post', $user->id), 403, 'Invalid permissions for this action');
 
         if($user->last_active_at == null) {
             return [];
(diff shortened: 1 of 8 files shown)


ThisIsMissEm:
You might find this video interesting: https://www.youtube.com/watch?v=Kf8SWqljM18

In Mastodon, user roles are stored as a bitmask, there is a "user roles" table, but it's a mapping between a role name and a bitmask, and then user's just have a column that references a role in that table. You could also use a json column instead for the role permissions too, instead of a bitmask.

That is it's Many Users to one UserRole, instead of a Many to Many relationship.

This has the benefit that's it's really easy to audit the system and see exactly who has what role, without doing a complex join, and also simplifies the administration side by having only a fixed set of roles, instead of many roles which may vary between users.

ThisIsMissEm:
That is, you could probably have a structure like this instead:

```php
        Schema::create('user_roles', function (Blueprint $table) {
            $table->id();
            $table->string('name')
            $table->json('permissions')->default(new Expression('(JSON_ARRAY())'));;
            $table->timestamps();
        });

        Schema::table('users', function (Blueprint $table) {
          $table->foreignId('role_id')->constrained('user_roles');
        });
```

Which would make it more easy for an administrator to assign roles to users since they'd have predefined categories that can do certain actions.

Also note the default value for the json column: This ensures that you're always working with an empty array, and not a null value.

app/Services/UserRoleService.php
```diff
@@ -0,0 +1,111 @@
...
+        $roleData = collect(self::roleData())
+            ->map(function($role, $k) use($myRoles) {
+                $role['value'] = $myRoles[$k];
+                return $role;
+            })
+            ->toArray();
+        return $roleData;
+    }
+
+    public static function roleData()
```
ThisIsMissEm, line 70: This looks like it should likely be handled by localisation on the frontend / API?

app/Services/UserRoleService.php
```diff
@@ -0,0 +1,111 @@
...
+
+        return self::defaultRoles();
+    }
+
+    public static function roleKeys()
+    {
+        return array_keys(self::defaultRoles());
+    }
+
+    public static function defaultRoles()
```
ThisIsMissEm, line 37: Following the comment around having a fixed set of roles, this would become your "default role" assigned to users (in Mastodon it has the `role_id` as nullable, and infers the default this way, as there's a "-99" default role created: https://github.com/mastodon/mastodon/blob/46bd58f74d11591a0180319285b0c79b2212ef69/app/models/user_role.rb#L105 )

app/Http/Controllers/Api/ApiV1Controller.php
```diff
@@ -1623,6 +1624,8 @@ public function mediaUpload(Request $request)
         ]);
 
         $user = $request->user();
+        abort_if($user->has_roles && !UserRoleService::can('can-post', $user->id), 403, 'Invalid permissions for this action');
```
ThisIsMissEm, line 1629: As every user would have the "default" role, you could remove the `$user->has_roles && ` here and just simplify to `abort_if(!UserRoleService::can('can-post', $user->id), 403, 'Invalid permissions for this action');`

This could also maybe be simplified through the usage of policies and gates: https://laravel.com/docs/10.x/authorization

dansup: We are using Policies and Gates, UserRoleService is just a cached wrapper around those!

app/Services/UserRoleService.php
dansup: I will add the l18n to the `getRoles` method.

Our frontend l18n is based on the backend Laravel language files in `resources/lang/{lang}` which does have Crowdin support (https://translate.pixelfed.org/), so share a single source of l18n files across backend and frontend!

app/Services/UserRoleService.php
dansup: I just pushed an updated defaultRoles!

ThisIsMissEm:
So given this change, what happens if the user is marked as not having roles, yet roles have been assigned, or vice-versa?

PR #4844, opened 2023-12-28
Refactor Docker/Compose

jippi:
Hello friends!

## Intro

This is a Request for Feedback (RFC) for improving the Docker experience with Pixelfed! 

It's a rather large amount of changes, but I've done my best to document inline why/what things are there for - both to aid in review and for future maintainers/consumers of it. 

As an active OSS maintainer myself, I know first hand that big "fly-by" improvements like this can be met with 🙄🙄🙄 and deep *sighs* for a lot of valid reasons (such as lack of testing, documentation, explanations, etc.) so I've tried my best to make everything relatively accessible and documented from the get-go. 

You can [read the explanation for a lot of the changes (and the documentation/usage) on my fork branch](https://github.com/jippi/pixelfed/tree/jippi-fork/docker#readme)

## Motivation

The motivation for this PR was to improve my own day-to-day experience managing my pixelfed server at https://pixelfed.dk - which is back by Docker and Docker Compose. 

I work with big scale platforms and infrastructure in my day job, so when I see an opportunity to improve the Docker experience for myself and others in OSS, I usually jump on it. 

## Outcomes of the changes

* Faster Docker builds thanks to  ...
  * multiple layers of build/layer caches ([~5-10min on warm cache on GitHub Actions](https://github.com/jippi/pixelfed/actions) - yes, even for `arm64`)
  * parallel [build stages](https://docs.docker.com/build/guide/multi-stage)
  * architecture (`amd64` and `arm64`) specific caches - they no longer overwrite each other
  * cached `apt` / `pecl` archives and binaries between runs
* Faster `docker pull` thanks to smaller, more efficient, layers (especially true when pulling updates!)
* Flexible Dockerfile via [build arguments](https://docs.docker.com/build/guide/build-args/) making customization easy and safe for server owners/consumers without having to hard-fork the Dockerfile
* Easy maintenance of software versions thanks to `build arguments` for PHP, Composer, and Nginx versions
* Merging of `fpm` and `apache` Dockerfiles for easier maintenance 
  * These are now [build targets](https://docs.docker.com/build/guide/multi-stage/#build-targets)
* Introduction of a `nginx` (Nginx + FPM in one container) runtime [build target](https://docs.docker.com/build/guide/multi-stage/#build-targets)
* [Smaller Docker images](https://hub.docker.com/r/jippi/pixelfed/tags) 
[...]

PR #4878, opened 2024-01-23
Refactor Docker/Compose from @jippi

Murazaki:
From @jippi PR in upstream : https://github.com/pixelfed/pixelfed/pull/4844

I had to squash for convenience (110 commits) and rebase on latest.

(If there's any concern on using latest staging please reply and tell me why)


Murazaki:
Again sorry, gonna get used to it eventually

PR #4891, opened 2024-02-02
Add S3 IG Import Media Storage

dansup:
Allows admins to opt-in to storing IG imported media on S3/cloud storage filesystem driver.

If you have Cloud Storage enabled, this will honor that setting and migrate media to S3 and delete the local copy.

### Configuration

Set `PF_IMPORT_IG_CLOUD_STORAGE=true` or `PF_IMPORT_IG_CLOUD_STORAGE=false` to enable or disable, keep in mind that you cannot enable this unless you have enabled Cloud Storage `PF_ENABLE_CLOUD=true`, however, you can disable this and keep IG import media stored locally (even when Cloud Storage is enabled) by setting this to `false`.

### Existing local media

Admins who update and have this enabled are not required to take any action, the cron job will be called automatically and will migrate existing and new IG media accordingly. It may take a few hours or days to process existing imports if you have a lot, but rest assured we thought of this. 

While we opted to silently update media urls so we don't send `Update` activities, you need not worry as local media urls will gracefully redirect to their S3 url when appropriate!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Added User Domain Blocks ([#4834](https://github.com/pixelfed/pixelfed/pull/4834)) ([fa0380ac](https://github.com/pixelfed/pixelfed/commit/fa0380ac))
 - Added Parental Controls ([#4862](https://github.com/pixelfed/pixelfed/pull/4862)) ([c91f1c59](https://github.com/pixelfed/pixelfed/commit/c91f1c59))
 - Added Forgot Email Feature ([67c650b1](https://github.com/pixelfed/pixelfed/commit/67c650b1))
+- Added S3 IG Import Media Storage support ([#4891](https://github.com/pixelfed/pixelfed/pull/4891)) ([081360b9](https://github.com/pixelfed/pixelfed/commit/081360b9))
 
 ### Federation
 - Update Privacy Settings, add support for Mastodon `indexable` search flag ([fc24630e](https://github.com/pixelfed/pixelfed/commit/fc24630e))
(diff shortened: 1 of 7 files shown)


ThisIsMissEm:
I'm still unsure of why you'd want to opt-in to storing all media in S3, unless it happens to have been imported from instagrams archives?

Having a single option to configure where to store media would be simpler, especially if it's all ultimately the same type of media: it just has a different source (user upload vs user import).

Can you explain to me why you'd want to have user media partially stored on disk and partially stored in S3?

dansup:
> I'm still unsure of why you'd want to opt-in to storing all media in S3, unless it happens to have been imported from instagrams archives?
> 
> Having a single option to configure where to store media would be simpler, especially if it's all ultimately the same type of media: it just has a different source (user upload vs user import).
> 
> Can you explain to me why you'd want to have user media partially stored on disk and partially stored in S3?

This is not opt-in, it is opt-out if you have cloud storage enabled.

ThisIsMissEm:
> > I'm still unsure of why you'd want to opt-in to storing all media in S3, unless it happens to have been imported from instagrams archives?
> > Having a single option to configure where to store media would be simpler, especially if it's all ultimately the same type of media: it just has a different source (user upload vs user import).
> > Can you explain to me why you'd want to have user media partially stored on disk and partially stored in S3?
> 
> This is not opt-in, it is opt-out if you have cloud storage enabled.

I'd probably recommend just having consolidated storage, and having the "move media to cloud" migration script run for all previous local instagram media and move that to cloud storage. i.e., discourage splitting media between storage backends, and encourage consolidating it.

PR #4921, opened 2024-02-11
Move CI to GitHub Actions

jippi:
👋 Hello friends,

This PR moves CI from Circle CI to GitHub Actions and introduces a couple of new steps

- `db` tests that migrations apply cleanly against supported databases
  - `mariadb:latest`
  - `mariadb:11`
  - `mariadb:10`
  - `mysql:latest`
  - `mysql:8.3`
  - `mysql:8.0`
  - `postgres:16`
  - `postgres:15`
  - `postgres:14`
- `php` tests that 
  - Supported PHP version (8.2 + 8.3) installs, builds, and tests passes cleanly
  - All files pass `php -l` (syntax check)
  - Tests all PHP files for code style against https://laravel.com/docs/10.x/pint
  - Uses same PHP image (https://hub.docker.com/_/php) and tooling as #4844 to keep it consistent   
  
Also lays the foundation for future tests using the various databases directly (or at all) in the future.

Further more, using GitHub Actions makes it easier for forks (since they likely also use GItHub) and it makes the testing output and status much more accessible to contributors directly, without going to 3rd party tools.

There are many opportunities to expand and improve our CI experience, having it managed in GitHub Actions will be a first important step for that.

### Questions

1. Do we want to use [Pint](https://laravel.com/docs/10.x/pint) for code style? Something else? nothing at all?
  - It seems to fail *a lot* currently https://github.com/jippi/pixelfed/actions/runs/7863233864/job/21453616057#step:10:1
  - `    FAIL   ............................ 1158 files, 1 error, 1000 style issues  `
  - Example PR applying pint to Pixelfed can be found here: https://github.com/pixelfed/pixelfed/pull/4922


composer.json
```diff
@@ -27,6 +27,7 @@
 		"laravel/passport": "^11.0",
 		"laravel/tinker": "^2.0",
 		"laravel/ui": "^4.2",
+		"lcobucci/clock": "^3.2",
```
jippi, line 30: Seems like 3.1 added PHP 8.2 support and 3.2 added PHP 8.3 supporting

With basically nothing else happening between those releases otherwise

.github/workflows/php.yml
```diff
@@ -0,0 +1,98 @@
...
+      group: php-test-${{ github.ref }}-${{ matrix.php_version }}
+      cancel-in-progress: false
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      # NOTE: we lint PHP files before composer to avoid wasting time
+      # linting everything in vendor/
+      - name: Lint PHP files
```
ThisIsMissEm, line 51: Would probably be good to split this out to a separate "lint / prettier" workflow to enforce code formatting.

That was something that made the recent security patches troublesome.

jippi: I'm fine with that - is there any specific `prettier` we want to use? 

This step only check them for being syntax correct PHP, it doesn't care about the structure or style itself. 

I put it here, very early in the job flow, to quickly get feedback before doing any other heavy lifting, but should be easy to move to a different job.

ThisIsMissEm: Ah, okay, so it's not so much a lint step, but a validation step before doing more expensive checks? I'd say fine to leave in here then

jippi: Yep, let me change it to use long form flag for more readability - it's mostly there to catch bad syntax or cross-PHP version incompatibilities (e.g. new functions/methods/classes)

```
--syntax-check / -l Syntax check only (lint)
```

jippi: I added https://laravel.com/docs/10.x/pint as a different job - I couldn't figure out what style guide PF uses, seems like "none", so picked Laravels own formatter :)

jippi: The PR show-casing pint applied to Pixelfed is here: https://github.com/pixelfed/pixelfed/pull/4922

.editorconfig
```diff
@@ -1,27 +1,25 @@
...
 
 [*.{yml,yaml}]
-indent_style = space
 indent_size = 2
 
 [*.{sh,envsh,env,env*}]
-indent_style = space
 indent_size = 4
 
 # ShellCheck config
```
jippi, line 17: this is just sorting keys alphabetically

.github/workflows/db.yml
```diff
@@ -0,0 +1,256 @@
...
+  workflow_dispatch:
+
+  # See: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push
+  push:
+    branches:
+      - '*'         # matches every branch that doesn't contain a '/'
+      - '*/*'       # matches every branch containing a single '/'
+      - '**'        # matches every branch
+    tags:
+      - "*"
```
ThisIsMissEm: I think you could just do `push:` here, which would match all pushes?

.github/workflows/db.yml
```diff
@@ -0,0 +1,256 @@
...
+        with:
+          repository: jippi/dottie
+          latest: true
+          fileName: "*_amd64.deb"
+
+      - name: Install dottie
+        run: |
+          dpkg -i *.deb
+          rm -f *.deb
+          dottie -v
```
ThisIsMissEm, line 105: Do you need the deb, or can you just download the binary?

.github/workflows/db.yml
```diff
@@ -0,0 +1,256 @@
...
+          key: composer-${{matrix.php_version}}-${{ hashFiles('composer.lock') }}
+
+      - name: Install composer dependencies
+        run: php composer.phar install -n --prefer-dist
+
+      - name: Setup Environment
+        run: |
+          cp .env.testing .env
+
+          dottie set DB_CONNECTION=mysql
```
ThisIsMissEm, line 125: Shouldn't this be `${{matrix.db}}` but stripping the tag from it?

.github/workflows/db.yml
```diff
@@ -0,0 +1,256 @@
...
+          dottie set REDIS_HOST=redis
+
+      - run: php artisan config:cache
+      - run: php artisan route:clear
+      - run: php artisan storage:link
+      - run: php artisan key:generate
+
+      - run: php artisan migrate --force
+
+  migrations-postgresql:
```
ThisIsMissEm, line 141: I think these can be merged? They do the same thing?

.github/workflows/db.yml
```diff
@@ -0,0 +1,256 @@
...
+      - name: Install composer dependencies
+        run: php composer.phar install -n --prefer-dist
+
+      - name: Setup Environment
+        run: |
+          cp .env.testing .env
+
+          dottie set DB_CONNECTION=pgsql
+          dottie set DB_HOST=db
+          dottie set DB_PORT=5432
```
ThisIsMissEm, line 239: Only these lines seem to change between the two.

.github/workflows/php.yml
```diff
@@ -0,0 +1,148 @@
...
+      - reopened
+      - synchronize
+
+env:
+  EXTRA_PHP_EXTENSIONS: intl bcmath zip pcntl exif curl gd pdo_pgsql pdo_mysql pdo_sqlite
+  DOCKER_PHP_EXTENSION_INSTALLER_VERSION: 2.2.2
+  COMPOSER_VERSION: 2.7.1
+
+jobs:
+  test:
```
ThisIsMissEm, line 24: I'd probably recommend splitting this into a lint and test jobs, such that you can fail the workflow fast on linting issues? Though you are using `fail-fast` false below?

.github/workflows/db.yml
jippi: there is a couple of differences between the MySQL and PostgreSQL jobs - and while it can be merged, I think the complexity of that isn't worth taking on vs a more verbose, but simple-to-follow step. CI should be as dumb as possible to keep it easy to debug and flexible.

.github/workflows/db.yml
jippi: there is a couple of differences between the MySQL and PostgreSQL jobs - and while it can be merged, I think the complexity of that isn't worth taking on vs a more verbose, but simple-to-follow step. CI should be as dumb as possible to keep it easy to debug and flexible.

* Different "health checks"
* Different DB "environment keys" (not PF ones, but the DB container image)

.github/workflows/db.yml
jippi: The deb is just the binary + 2 shell completion scripts, and it ensures the file ends up in the right path and so on, so difference is ~1-2 KB and the deb file is simpler/more flexible going forward.

.github/workflows/php.yml
jippi: I merged the two into for a couple of reasons:

1. Avoiding confusing "cause and effect" situations where a syntax error caused many pipelines to fail (and having to figure out which) - or tests failing under PHP 8.x fail but not on PHP 8.y due to some syntax/issue.
2. There is no point in running tests (or its pipeline) if there is a syntax error, so it shortens full pipeline runtime from end-to-end.
3. Fewer CI jobs kicking off at the same time, to keep the number of CI jobs ~low. 

I'm going to pull out the `pint` part for now though, as @dansup seem to want to do this a different way / timing.

PR #4956, opened 2024-02-24
buzz/laravel-h-captcha: allow semantic range to make `composer validate` happy

LeSuisse:
Currently `composer validate` complains about the exact version constraint on this package. Looking at the `git log` the constraint does not seem to exist for a specific reason.

No functional changes expected.

```
$ composer validate
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
- require.buzz/laravel-h-captcha : exact version constraints (1.0.4) should be avoided if the package follows semantic versioning
```


dansup:
Fixed in 5fb26a78

PR #4962, opened 2024-02-29
Add Remote Reports to Admin Dashboard Reports page

dansup:


app/Http/Controllers/Admin/AdminReportController.php
```diff
@@ -665,6 +669,24 @@ public function reportsApiAll(Request $request)
...
+        $filter = $request->input('filter') == 'closed' ? 'closed' : 'open';
+
+        $reports = AdminRemoteReport::collection(
+            RemoteReport::orderBy('id', 'desc')
+                ->when($filter, function ($q, $filter) {
+                    return $filter == 'open' ?
+                    $q->whereNull('action_taken_at') :
+                    $q->whereNotNull('action_taken_at');
+                })
+                ->cursorPaginate(6)
```
ThisIsMissEm, line 683: Showing only 6 at a time seems like an odd choice, considering some instances may be receiving hundreds of reports a day (this was the case for us at switter.at)

app/Http/Controllers/Admin/AdminReportController.php
```diff
@@ -1327,4 +1349,173 @@ public function reportsApiSpamGet(Request $request, $id)
...
+
+        $report = RemoteReport::findOrFail($request->input('id'));
+        $user = User::whereProfileId($report->account_id)->first();
+        $ogPublicStatuses = [];
+        $ogUnlistedStatuses = [];
+        $ogNonCwStatuses = [];
+
+        switch ($request->input('action')) {
+            case 'mark-read':
+                $report->action_taken_at = now();
```
ThisIsMissEm, line 1368: There is no action actually being taken here? Why isn't this just "close report"?

app/Http/Controllers/Admin/AdminReportController.php
```diff
@@ -1327,4 +1349,173 @@ public function reportsApiSpamGet(Request $request, $id)
...
+                $report->save();
+                break;
+            case 'mark-all-read-by-domain':
+                RemoteReport::whereInstanceId($report->instance_id)->update(['action_taken_at' => now()]);
```
ThisIsMissEm, line 1372: If you really want to ignore all reports from a domain, which should really be a last-resort action, then you'd want to handle this via Federation Management, whereby you reject all incoming reports from that instance.

it may make sense to be able to filter reports by instance, such that you can have a queue to easily process. The side effect of having a bulk action like this is that you might throw out a good report with all the bad reports

app/Http/Controllers/Admin/AdminReportController.php
```diff
@@ -1327,4 +1349,173 @@ public function reportsApiSpamGet(Request $request, $id)
...
+                break;
+            case 'cw-posts':
+                $statuses = Status::find($report->status_ids);
+                foreach($statuses as $status) {
+                    if($report->account_id != $status->profile_id) {
```
ThisIsMissEm, line 1377: This filter should happen on the receiving side I think? Keep in mind Mastodon will send you Flag activities where your user is not the reported user, but rather a user mentioned in that reply.

app/Http/Controllers/Admin/AdminReportController.php
```diff
@@ -1327,4 +1349,173 @@ public function reportsApiSpamGet(Request $request, $id)
...
+                    StatusService::del($status->id);
+                }
+                break;
+            case 'private-posts':
+                $statuses = Status::find($report->status_ids);
+                foreach($statuses as $status) {
+                    if($report->account_id != $status->profile_id) {
+                        continue;
+                    }
+                    if(in_array($status->scope, ['public', 'unlisted', 'private'])) {
```
ThisIsMissEm, line 1438: This check is redundant given the check on the next line as far as I can tell?

PR #5002, opened 2024-03-10
Docker: Script to check requirements

jippi:
To help folks with their Docker setup, this script will provide automatic checks (and actionable errors) for the most common issues so observed so far.

Particularly around Docker Compose being too old, resulting in semi-confusing errors such as: 

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.db.ports contains an invalid type, it should be a number, or an object
services.redis.ports contains an invalid type, it should be a number, or an object
```

This Docker Compose version of `2.17` is *arbitrary*, as I do not know what the _actual_ minimum version is, but it was selected based on being a [one year old release](https://docs.docker.com/compose/release-notes/#2170) and with the expectation that woulds should keep their system and software somewhat up to date, as Docker has patched many CVEs in both the Docker engine and Docker Compose since `2.17`

A successful check will look like this
Image 2024-03-10 at 4 01 10 PM jpg

Where a failed "minor version" check for compose will look like this
Image 2024-03-10 at 4 01 46 PM jpg

And finally, a "major version" check fail for compose looks like this
Image 2024-03-10 at 4 02 22 PM jpg


jamespanic:
Hi I tried the script and got the following but I still receive the errors mentioned above: 

```
👋 Hello!

This script will check your system for the minimum requirements outlined in the Pixelfed Docker install guide
You can find the guide here https://jippi.github.io/pixelfed-docs-next/pr-preview/pr-1/running-pixelfed/docker/prerequisites.html#software.

⚙️ Checking if [git] command is available:
        ✅ git is installed

⚙️ Checking if [docker] command is available:
        ✅ docker is installed

⚙️ Checking if [docker compose] command is available:
        ✅ docker compose is installed

⚙️ Checking if [docker compose version] major version (2) is acceptable:
        ✅ You're using major version 2

⚙️ Checking if [docker compose version] minor version (17) is acceptable:
        ✅ You're using minor version 24

🎉 All checks passed, you should be ready to run Pixelfed on this server!
```

```
❯ docker-compose version
docker-compose version 1.29.2, build unknown
docker-py version: 5.0.3
CPython version: 3.11.2
OpenSSL version: OpenSSL 3.0.11 19 Sep 2023
```

Maybe I need to be running `docker compose` instead of `docker-compose` ?

jippi:
Nice catch @jamespanic - going to add a check to ensure `docker compose` and `docker-compose` should give same version 🤔

ricardo-maestas:
The old Docker compose example recently tripped me up. This would have been very helpful!

dansup (approved):

PR #5007, opened 2024-03-14
fix: Unable to delete server rules

### LINKED ISSUE #4990: Unable to delete server rules
Pixelfed 0.11.13

When attempting to delete a rule, the server responds with an error saying that regs is required . Look like an invalid validation for the delete rule rest route.

The request to delete is only sent with `{"rule_delete":"2"}`

image

MuhammadSaim:
When attempting to delete a rule, the server responds with an error saying that regs is required . Look like an invalid validation for the delete rule rest route.

The request to delete is only sent with <kbd>{"rule_delete":"2"}</kbd>

image

Fixes: #4990 

For this PR I just change the rule for <kbd>regs</kbd> which is required and change that to <kbd>nullable</kbd>. Verify it from the frontend also only the <kbd>rule_delete</kbd> is passing in post parameters.

### DIFF
--- a/app/Http/Controllers/Admin/AdminSettingsController.php
+++ b/app/Http/Controllers/Admin/AdminSettingsController.php
@@ -69,7 +69,7 @@ trait AdminSettingsController
             'type_mp4' => 'nullable',
             'type_webp' => 'nullable',
             'admin_account_id' => 'nullable',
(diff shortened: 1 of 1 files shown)


dansup:
Fixed in #5005

PR #5019, opened 2024-03-28
Add support for redis username & url configuration; fixes #4909

### LINKED ISSUE #4909: Make username for redis connection configureable
Currently redis authentication in pixelfed is only possible via the default user. It would be good if a username can be specified for people who want to utilize redis ACL.

Tested it with following patch:

```diff
--- config/database.php.orig	2024-02-07 20:10:49.297463309 +0100
+++ config/database.php	2024-02-07 20:11:17.565973954 +0100
@@ -118,6 +118,7 @@
             'scheme'   => env('REDIS_SCHEME', 'tcp'),
             'path'     => env('REDIS_PATH'),
             'host'     => env('REDIS_HOST', '127.0.0.1'),
+            'username' => env('REDIS_USERNAME', null),
             'password' => env('REDIS_PASSWORD', null),
             'port'     => env('REDIS_PORT', 6379),
             'database' => env('REDIS_DATABASE', 0),
```

```diff
--- config/cache.php.orig	2024-02-07 20:13:02.027860986 +0100
+++ config/cache.php	2024-02-07 20:13:22.968239270 +0100
@@ -76,6 +76,7 @@
                 'scheme'   => env('REDIS_SCHEME', 'tcp'),
                 'path'     => env('REDIS_PATH'),
                 'host'     => env('REDIS_HOST', 'localhost'),
+                'username' => env('REDIS_USERNAME', null),
                 'password' => env('REDIS_PASSWORD', null),
                 'port'     => env('REDIS_PORT', 6379),
                 'database' => env('REDIS_DATABASE', 0),
```

Mai-Lapyst:
Edit: link to issue #4909

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -77,9 +77,11 @@ return [
             'client' => env('REDIS_CLIENT', 'predis'),
 
             'default' => [
+                'url'      => env('REDIS_URL', null),
                 'scheme'   => env('REDIS_SCHEME', 'tcp'),
                 'path'     => env('REDIS_PATH'),
                 'host'     => env('REDIS_HOST', 'localhost'),
+                'username' => env('REDIS_USERNAME', null),
                 'password' => env('REDIS_PASSWORD', null),
                 'port'     => env('REDIS_PORT', 6379),
                 'database' => env('REDIS_DATABASE', 0),
--- a/config/database.php
+++ b/config/database.php
@@ -115,9 +115,11 @@ return [
         'client' => env('REDIS_CLIENT', 'predis'),
 
         'default' => [
+            'url'      => env('REDIS_URL', null),
             'scheme'   => env('REDIS_SCHEME', 'tcp'),
             'path'     => env('REDIS_PATH'),
             'host'     => env('REDIS_HOST', '127.0.0.1'),
+            'username' => env('REDIS_USERNAME', null),
             'password' => env('REDIS_PASSWORD', null),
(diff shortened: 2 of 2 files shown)


x64x2 (commented):
Is this really necessary??

Mai-Lapyst:
> Is this really necessary??

@ThisIsMissEm said it would be a reasonable addition and I agree. Atleast for the `username` setting, as I want to configure my redis with tight user permissions as is possible for quite some time now. But this requires that every application I use the redis server with needs to login with a unique username and not relay on the backwards compatible way of utilising the default user (i.e. not sending a username to begin with).

ThisIsMissEm (commented):
I think it should be URL or the individual options, I don't think both work together?

Mai-Lapyst:
> I think it should be URL or the individual options, I don't think both work together?

Im not that familiar with the depths of laravel, but atleast the homepage shows examples where they too use the url and the individual options in one file: https://laravel.com/docs/11.x/redis#configuring-the-connection-scheme so I think it should work? Tested it on my instance and it worked flawlessly.

dansup:
To ensure backwards compatibility with regards to configuration, I will not be merging this.

You can edit the `config/*.php` config files as needed, a better solution would be adding this to our docs.

ThisIsMissEm:
@dansup I tested this and it would be backwards compatible as far as I can tell, because if these properties are null, then they are not used by the redis client binding.

PR #5021, opened 2024-03-29
Implement Mastodon Admin Domain Blocks API

ThisIsMissEm:
This is an API that provides compatibility with the Mastodon API for Admin Domain Blocks, whilst still using the underlying pixelfed setup of storing this information on the `Instances` table.


ThisIsMissEm:
I think this is now done. @dansup what's your preference around testing?

app/Http/Controllers/Api/V1/Admin/DomainBlocksController.php
```diff
@@ -0,0 +1,146 @@
...
+    if ($parts[0] == '*') {
+      // If we only have two parts, e.g., "*", "example", then we want to fail:
+      abort_if(count($parts) <= 2, 400, 'Invalid domain: This API does not support wildcard domain blocks yet');
+
+      // Otherwise we convert the *.foo.example to foo.example
+      $domain = implode('.', array_slice($parts, 1));
+    }
+
+    // Double check we definitely haven't let anything through:
+    abort_if(str_contains($domain, '*'), 400, 'Invalid domain');
```
ThisIsMissEm: I know this is nasty, but couldn't think of a better way to do it. Mastodon's domain block functionality supports these wildcard blocks, e.g., `*.example` or `*.social.example`, in Pixelfed, as we're working with instances, we need to find something that could plausibly be an instance, so `example` can't be, but `social.example` could be.

This will be greatly improved if we get the NLNet funding to overhaul the domain federation management.

app/Http/Controllers/Api/V1/Admin/DomainBlocksController.php
```diff
@@ -0,0 +1,147 @@
...
+    $this->validate($request, [
+      'limit' => 'sometimes|integer|max:100|min:1',
+    ]);
+
+    $limit = $request->input('limit', 100);
+
+    $res = Instance::moderated()
+      ->orderBy('id')
+      ->cursorPaginate($limit)
+      ->withQueryString();
```
ThisIsMissEm, line 29: We may want to add caching here, but tbh, it's not necessary due to this being an administrative API.

app/Providers/AuthServiceProvider.php
```diff
@@ -24,7 +24,7 @@ class AuthServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        if(config('app.env') === 'production' && config('pixelfed.oauth_enabled') == true) {
+        if(config('pixelfed.oauth_enabled') == true) {
```
ThisIsMissEm: This fixes the OAuth Applications UI not working in development — as long as oauth_enabled is true, you should be able to use the OAuth API, regardless of the environment.

ThisIsMissEm: @dansup aside, do we disable the Mastodon API if OAuth isn't enabled?

app/Http/Controllers/Api/V1/Admin/DomainBlocksController.php
```diff
@@ -0,0 +1,147 @@
...
+use App\Http\Controllers\Api\ApiController;
+use App\Instance;
+use App\Services\InstanceService;
+use App\Http\Resources\MastoApi\Admin\DomainBlockResource;
+
+class DomainBlocksController extends ApiController {
+
+  public function __construct() {
+    $this->middleware(['auth:api', 'api.admin', 'scope:admin:read,admin:read:domain_blocks'])->only(['index', 'show']);
+    $this->middleware(['auth:api', 'api.admin', 'scope:admin:write,admin:write:domain_blocks'])->only(['create', 'update', 'delete']);
```
ThisIsMissEm, line 16: @dansup this is probably how we should have done the Access Token verification checks in the security patch. I couldn't test at the time so went with the "simpler" to validate option.

app/Http/Controllers/Api/ApiController.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace App\Http\Controllers\Api;
+
+use Illuminate\Http\Request;
+use App\Http\Controllers\Controller;
+
+class ApiController extends Controller {
```
ThisIsMissEm, line 8: Inheriting from this `ApiController` class will allow us to delete a LOT of code, because we've a lot of repeated code for doing JSON and the link headers.

app/Http/Controllers/Api/V1/Admin/DomainBlocksController.php
```diff
@@ -0,0 +1,147 @@
...
+    return $this->json(DomainBlockResource::collection($res), [
+      'Link' => $this->linksForCollection($res)
+    ]);
+  }
+
+  public function show(Request $request, $id) {
+    $domain_block = Instance::moderated()->find($id);
+
+    if (!$domain_block) {
+      return $this->json([ 'error' => 'Record not found'], [], 404);
```
ThisIsMissEm, line 40: Might do a `$this->notFound()` helper too.

dansup:
@ThisIsMissEm I closed this because I noticed a few issues, however, I will re-open this because you have done some great work here and it's worth polishing up to merge. I'm mostly concerned regarding the removal of the env check, and adding the api middleware to the controller, I'm not sure that would work with web requests but I will test and do a full review.

app/Providers/AuthServiceProvider.php
dansup: Nope, we should make OAuth required as the webUI and official web app use it

dansup (approved):
Nice work!

PR #5096, opened 2024-05-19
Update DirectMessageController, add 72 hour delay for new accounts before they can send media directly

havijs:
I was trying to send a text message along with a photo via DM. The photo was sent, but the text resulted in a 400 error. I think both text and media should follow the same rule.


dansup:
Added in 4aad1c22

PR #5097, opened 2024-05-19
Fix environment variables

intentionally-left-nil:
When a bash array is read into a variable, the syntax must be read -r -a <variable name> not read -ar This fixes ENTRYPOINT_SKIP_SCRIPTS as well as DOCKER_APP_ENSURE_OWNERSHIP_PATHS

Currently, if you try to set the ENTRYPOINT_SKIP_SCRIPTS variable, you'll see that nothing is skipped. From debugging this is because the skip_scripts array is never initialized


intentionally-left-nil:
FYI you can play around with these fixes with `docker pull ghcr.io/intentionally-left-nil/pixelfed-fpm:0.12` and see that it works for your setups

dansup (changes requested):

dansup (approved):

PR #5098, opened 2024-05-19
If not applying migrations, early exit before waiting for the database

intentionally-left-nil:
If we're not going to apply the migrations, there's no point in waiting for the database to be available, just quick-start the service.

Otherwise there's no way to opt-out of `await-database-ready` without skipping every single entrypoint file


intentionally-left-nil:
On second thought, you can use ENTRYPOINT_SKIP_SCRIPTS to skip this whole file if you want. I don't think this change is needed

PR #5099, opened 2024-05-19
Fix parsing multiline variables

intentionally-left-nil:
The seen_dot_env_variables parser does not work correctly for multiline variables Instead, just use set -o allexport to perform the needed behavior in the template file

To repro this:
create a .env file that has a multiline variable in it
```
myvariable="line1
line2"
```

and then try to run the docker container. The templating file will fail because seen_dot_env_variables is wrong (the grep is per-line)

The fix here is to use set -o allexport and just let bash do it for us


intentionally-left-nil:
FYI you can play around with these fixes with `docker pull ghcr.io/intentionally-left-nil/pixelfed-fpm:0.12`

dansup (changes requested):

dansup (approved):

PR #5191, opened 2024-06-28
Extended translation coverage in the home/personal feed page

whitenoise2:
Hello Pixelfed;

This is my first attempt at extending translation coverage (I am bilingual French/English, so French is my translated language) - I've extended the translation on the Home feed UI page, and I wanted to share the results to see if it would be something the project would be interested in.

I've re-jigged some of the existing translation, before I became aware of the translation site setup - my version is aligned with the Quebec government's official tech translation guide; I will totally revert back to the previous values if that works better -- the idea is to keep on adding translation to Pixelfed until it is fully localised.

I understand that there are more challenges down the road (units, sentences that embed html, etc) but I've started reading more in-depth on translation templates and other Vue/Laravel-Illuminate tricks, so I think it will be feasible.


whitenoise2:
I could also move my efforts to fr-ca, in order to make it a french canadian translation, if this helps.

dansup:
Please send translations via https://translate.pixelfed.org/

whitenoise2:
Hi there;

How does one go and submit vue.js template changes through crowdin?

PR #5210, opened 2024-07-07
Mark RUNTIME_GID as an argument

intentionally-left-nil:
When building the dockerfile, there's currently this warning:
```
  1 warning found (use --debug to expand):
 - UndefinedVar: Usage of undefined variable '$RUNTIME_GID' (line 219)
Dockerfile:262
```

This PR fixes the issue


intentionally-left-nil:
Actually, this is a required fix.

If you set the build-arg `BUILD_FRONTEND=1` the current dockerfile won't build without this fix

jippi:
Do you mind reopening these against https://github.com/jippi/docker-pixelfed instead? as [mentioned on Discord](https://discord.com/channels/446875483016134658/1230743309496877066/1259501713065377843) thats where the ongoing maintenance and development of the Docker + Pixelfed experience is happening

intentionally-left-nil:
Copy-pasta :)

Just summarizing the conversation from discord, I'm not interested in contributing to a downstream project right now.

As long as the Dockerfile exists in this repository I would expect for the pixelfed git repo to accept changes to it. If that's not the case, then Dan should remove all of the code which is unsupported.

dansup (approved):

PR #5211, opened 2024-07-07
Docker fixes for composer install

intentionally-left-nil:
This implements two fixes: First, it copies over all source files before running composer install. This is needed because some post-install steps use artisan, which requires the source to be present Second, this fixes the permissions of /cache/composer. Otherwise, it is owned by root, and is not writeable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -256,16 +256,25 @@ COPY --link --from=composer-image /usr/bin/composer /usr/bin/composer
 #! Changing user to runtime user
 USER ${RUNTIME_UID}:${RUNTIME_GID}
 
+
 # Install composer dependencies
 # NOTE: we skip the autoloader generation here since we don't have all files avaliable (yet)
-RUN --mount=type=cache,id=pixelfed-composer-${PHP_VERSION},sharing=locked,target=/cache/composer \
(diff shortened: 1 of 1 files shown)


intentionally-left-nil:
ah actually this still isn't working. Let me keep working at it

intentionally-left-nil:
Okay, this is working on my image: https://github.com/intentionally-left-nil/pixelfed/pkgs/container/pixelfed-fpm/240184861?tag=0.12.3

and I'm happy with the testing

jippi:
Do you mind reopening these against https://github.com/jippi/docker-pixelfed instead? as [mentioned on Discord](https://discord.com/channels/446875483016134658/1230743309496877066/1259501713065377843) thats where the ongoing maintenance and development of the Docker + Pixelfed experience is happening

intentionally-left-nil:
Just summarizing the conversation from discord, I'm not interested in contributing to a downstream project right now.

As long as the Dockerfile exists in this repository I would expect for the pixelfed git repo to accept changes to it. If that's not the case, then Dan should remove all of the code which is unsupported.

intentionally-left-nil:
@dansup this PR has been sitting for awhile. Can you take a look

Casuallynoted:
Also curious about what the status of this is @dansup

dansup (approved):

PR #5212, opened 2024-07-07
Translation coverage extension and automated build process

whitenoise2:
This pull request is about MORE TRANSLATION in pixelfed, not just changing existing strings.

It contains:
* extended Vue.js templates with additional translation handling code - including use of singular/plural forms where appropriate - for the landing and Timeline views
* support for language variants in dropdowns (admin section)
* additional entries in web.php for english and French (Canadian) to cover the additions/serve as example
* additional keys and values where beneficial to non-english languages.

this work cannot be accomplished via crowdin, but adds to what is available at crowdin - a custom language variant for french (canada) will need to be added to crowdin (would be happy to help with that if possible).

Will be:
* normalizing all web.php files to have all strings with english value where not already defined
* add more translation coverage


dansup (changes requested):
Please remove all `public/` assets, those get generated automatically!

whitenoise2:
Woopsies!

Duly noted, will remove.

Currently also integrating phing as a php-based build framework to remove i18n actions from artisan.

Will update branch accordingly later this week.

whitenoise2:
Managed to get some more work done - added Phing support properly, with .json file generation moved from artisan command to Phing task.

Ensured that Laravel app is properly started and lives during the entire task lifetime, so that custom tasks can be written with the same ease of coding as artisan commands.

Added build bootstrap scripts (build.sh and build.bat) for build startup with DOS and *nix.

Wanting to add two more custom tasks:
* detect translations - scan php files and vue.js views for translation function calls and add missing keys into the english blueprint for translation
* normalise translations  - automagically propagate all missing keys from the english blueprint to all other languages, with english value supplied as default

these two tasks combined would auto-propagate code-defined translation keys to translation php maps.

regarding the app split - is Pixelfed using Vue.js/SPA only for the customer-facing app and keeping PHP for the rest? meaning that all SPA translations would be in web.php and the other translation files will stay? (if so, I will extend support for both -- I would then prefix the translation keys with the target file they reside in to enable automatic addition of newly detected keys to proper files)

I've also started using nested hashes to group translation keys over the previous inline comments technique - I find it easier to navigate and it helps with semantic descriptions of keys for translators.

then, of course, will come build tasks for retrieving the latest translation maps from crowdin and merging them in, and for pushing update translation maps to crowdin.

PR #5250, opened 2024-07-31
Laravel shift PR to upgrade pixelfed to Laravel 11

Synchro:
This is a direct share of the PR that I created with [Laravel Shift](https://laravelshift.com/) to upgrade pixelfed to Laravel 11.

Each shift feature is applied as a separate commit, so you can cherry-pick as you like.

Even if you don't use this directly (it makes *a lot* of changes!), it at least gives you a clear impression of what you will need to do for L11 compatibility.

It is very large, primarily because it also applies the Laravel coding standard and there are many lines that do not conform.

I hope you find it useful!


dansup:
@Synchro Hey, thanks for this but I did already update to v11 in https://github.com/pixelfed/pixelfed/commit/f7bbf3eab58eeeee5d38c76c8042c28d6c64972c

There are a lot of good changes that I didn't catch, so I will def be referencing this!

Cheers!

PR #5282, opened 2024-09-05
Fix Move inbox handler

dansup:
Fix Federation logic to process federated `Move` activities.

Will be compatible with Mastodon and other implementations that support the `Move` activity with `alsoKnownAs` and `movedTo` attributes.

[Spec](https://swicg.github.io/activitypub-data-portability/#process-0)

### DIFF
--- a/app/Jobs/MovePipeline/ProcessMovePipeline.php
+++ b/app/Jobs/MovePipeline/ProcessMovePipeline.php
@@ -3,6 +3,7 @@
 namespace App\Jobs\MovePipeline;
 
 use App\Services\ActivityPubFetchService;
+use App\Util\ActivityPub\Helpers;
 use DateTime;
 use Exception;
 use Illuminate\Contracts\Queue\ShouldQueue;
--- a/app/Jobs/MovePipeline/ProcessMovePipeline.php
+++ b/app/Jobs/MovePipeline/ProcessMovePipeline.php
@@ -61,7 +61,7 @@ class ProcessMovePipeline implements ShouldQueue
      */
     public function retryUntil(): DateTime
     {
-        return now()->addMinutes(15);
+        return now()->addMinutes(5);
     }
 
     /**
@@ -80,7 +80,7 @@ class ProcessMovePipeline implements ShouldQueue
         if (! self::checkActor()) {
             throw new Exception('Invalid actor');
         }
-
+        return;
     }
 
     protected function checkTarget()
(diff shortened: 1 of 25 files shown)


app/Jobs/MovePipeline/ProcessMovePipeline.php
```diff
@@ -0,0 +1,139 @@
...
+            Log::info('[AP][INBOX][MOVE] target fetch failure');
+            return false;
+        }
+
+        if (is_string($res['alsoKnownAs'])) {
+            return self::lowerTrim($res['alsoKnownAs']) === self::lowerTrim($this->actor);
+        }
+
+        if (is_array($res['alsoKnownAs'])) {
+            $map = array_map(self::lowerTrim(), $res['alsoKnownAs']);
```
cmack: Does this map work as expected? It looks like it's doing a full function call instead of passing `lowerTrim` as a callable to `array_map`

dansup: Fixed in 5ad25759

app/Jobs/MovePipeline/UnfollowLegacyAccountMovePipeline.php
```diff
@@ -0,0 +1,157 @@
...
+                $pool = new Pool($client, $requests($followers), [
+                    'concurrency' => config('federation.activitypub.delivery.concurrency'),
+                    'fulfilled' => function ($response, $index) {},
+                    'rejected' => function ($reason, $index) {},
+                ]);
+
+                $promise = $pool->promise();
+
+                $promise->wait();
+            }, 'id');
```
cmack: @dansup I know you're trying to get this feature out the door, and I'm new to this project (not to Laravel), but I want to ask constructively:
1. Would this logic be better in its own queued Job and the loop body a quick `dispatch(WellNamedJob::class);`
2. If any of the Guzzle requests fail, would it affect the chunking query as a whole? i.e would the rest of the move logic fail on the first follower failure? Would that be desireable?
3. I mention the separate job because they can be individually restarted via Horizon if they fail and that seems like it could be nice here, to my Laravel-eye.

Apologies, if this is already on your burndown list / radar.

dansup: @cmack Yeah that is a great idea, am finishing up the refactor as we speak!

dansup: Shipped in 15918cba!

PR #5344, opened 2024-11-12
Update AP helpers, reject statuses with invalid dates

dansup:


app/Util/ActivityPub/Helpers.php
```diff
@@ -329,6 +345,10 @@ public static function statusFirstOrFetch($url, $replyTo = false)
             return;
         }
 
+        if (! self::validateTimestamp($res['published'])) {
+            return;
+        }
```
ThisIsMissEm, line 350: You may also want to consider similar for updated at / created at.

PR #5345, opened 2024-11-12
internationalization of site/contact

nesges:
Added new keys for language used in resources/views/site/contact.blade.php and added them in resources/lang/en/site.php with the original text and resources/lang/de/site.php with a german translation


nesges:
closing here to send a new PR with target staging

PR #5361, opened 2024-11-24
Fix photo_reblogs_only not working

Murazaki:
This makes sure timeline fetches knows if a reblog is of a valid type, and shows it according to `photos_reblogs_only` setting.

You will need to rebuild UI assets for it to work.

### DIFF
--- a/app/Http/Controllers/Api/ApiV1Controller.php
+++ b/app/Http/Controllers/Api/ApiV1Controller.php
@@ -2469,6 +2469,7 @@ class ApiV1Controller extends Controller
             'max_id' => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
             'limit' => 'sometimes|integer|min:1',
             'include_reblogs' => 'sometimes',
+            'photos_reblogs_only' => 'sometimes',
         ]);
 
         $napi = $request->has(self::PF_API_ENTITY_KEY);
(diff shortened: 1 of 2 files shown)


x64x2 (commented):
Makes no changes.

Murazaki:
> Makes no changes.

if you want to test it, you will need to rebuild UI assets.

I guarantee you this changes how reblogs are shown, as I've been testing it on Pixelfed Glitch with other reviewers.

Murazaki:
Please review.

x64x2 (approved):

dansup:
@Murazaki Thanks for the PR, I will profile the database query changes this weekend and see if I can make any optimizations before merging.

This is one of the most complex queries we do, so I want to make sure we have all our bases covered. Expect an update in the next few days.

Murazaki:
> @Murazaki Thanks for the PR, I will profile the database query changes this weekend and see if I can make any optimizations before merging.
> 
> This is one of the most complex queries we do, so I want to make sure we have all our bases covered. Expect an update in the next few days.

Thank you, that's why I'm sharing it before merging on my side, I wanted to make sure this doesn't bloat the query.

weddige:
> Expect an update in the next few days.

Any updates to this? The fact that reblogs are broken (see also #5573) is the main reason preventing me from using Pixelfed more often.

Issue #5394, opened 2024-12-16
Video thumbnails missing from hashtag page


witten:
Video thumbnails are generated correctly, and they show up properly on both the home timeline and the profile page. But thumbnails for videos are missing from the hashtag page, appearing as just a blank off-white box there:

![nothumb](https://github.com/user-attachments/assets/eaf62709-e9e1-4383-babe-0478e745b305)

Looking at my browser console during page load, there are no JavaScript errors or HTTP 404s. There are HTTP GETs that return a 200 or 304. Interestingly, the video .mp4 files themselves are being requested on page load, which might indicate the problem.. Perhaps instead of video thumbnails being loaded (PNG or JPEG), the video files themselves are being requested:

![wrongurl](https://github.com/user-attachments/assets/9fb8015a-b1e2-4348-86f8-e2535c9a6f72)

Possibly related: #3402

dansup:
Fixed by your PR!

PR #5395, opened 2024-12-16
Fix missing video thumbnails on hashtag page (#5394).

witten:
**Changes**

This implements #5394. As described in that ticket, video thumbnails were missing from the hashtag page, and it appeared that the wrong URL was being requested. Upon closer inspection (looking at the DOM elements of the page), that did appear to be the case.. The `` for any video thumbnail ended in `.mp4` rather than the expected `.png` or `.jpg`. So switching the page from using just `.url` to `.preview_url` for each status' media attachments appeared to fix the problem—and now the correct video thumbnail shows up. Also, thumbnails for non-video media attachments continue showing up properly.

**Testing done**

I don't have a pipeline set up for Pixelfed source :arrow_right: container currently, so I hand-hacked the JavaScript of the bundle corresponding to the modified Vue file. :grimacing: Then I reloaded a hashtag page on my site and observed that video thumbnails loaded alongside the existing image thumbnails.

Let me know if there are any automated tests I should add here.. I didn't see existing tests for any Vue components.

**Future work**

It's beyond my abilities right now (I'm a backend dev), but now that there are working thumbnails, it would probably be a good idea to bring over the `video-overlay-badge` (and `timestamp-overlay-badge`?) from the profile page (`ProfileFeed.vue` ). That way, it'd be clear which thumbnail here corresponds to a video and which is just an image.

### DIFF
--- a/resources/assets/components/Hashtag.vue
+++ b/resources/assets/components/Hashtag.vue
@@ -74,7 +74,7 @@
 												width="32"
 												height="32"
 												:hash="status.media_attachments[0].blurhash"
-												:src="status.media_attachments[0].url"
+												:src="status.media_attachments[0].preview_url"
 												/>
 										</div>
 										<span v-if="status.pf_type == 'photo:album'" class="float-right mr-3 post-icon"><i class="fas fa-images fa-2x"></i></span>


dansup (changes requested):
Please target the staging branch

witten:
Will do. I'll open a new PR for that.

witten:
#5427

PR #5402, opened 2024-12-21
imagick instead of gd

blitzgneisserin:
Gd changes the colors of non srgb pics

### DIFF
--- a/config/image.php
+++ b/config/image.php
@@ -15,6 +15,6 @@ return [
     |
     */
 
-    'driver' => env('IMAGE_DRIVER', 'gd'),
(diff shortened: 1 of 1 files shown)


dansup:
Not every system has Imagemagick, and I'm in the process of implementing libvips support so we will default to that.

blitzgneisserin:
Please make sure that it's possible to upload non srgb content without changing the colors

dansup:
> Please make sure that it's possible to upload non srgb content without changing the colors

That will be possible soon, where supported. We need to keep GD as it would be a BC

PR #5439, opened 2025-01-05
Use DOCKER_PROXY_VERSION from .env file

LiohMoeller:
Use DOCKER_PROXY_VERSION from .env file
Opted for major version 1.6 instead of minor


dansup (changes requested):

dansup (approved):

PR #5442, opened 2025-01-06
Docker Registry

LiohMoeller:
Registry has changed. Old registry has been discontinued in August 2024. New Registry added, format of Docker tag has been adjusted as it now contains the Debian Release as well.

Sample Version is set to current stable but can be adjusted to any of the available branches.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -964,7 +964,7 @@ TZ="${APP_TIMEZONE}"
 # Combined with [DOCKER_APP_RUNTIME] and [PHP_VERSION] configured
 # elsewhere in this file, the final Docker tag is computed.
 # @dottie/validate required
-DOCKER_APP_RELEASE="v0.12.4"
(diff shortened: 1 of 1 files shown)


LiohMoeller:
Now I get an error during docker compose like: 👷 Running [dottie validate --file /var/www/.env --ignore-rule dir,file --exclude-prefix APP_KEY --no-fix] as [root]
pics.spicy.social-worker      | [entrypoint / 02-check-config.sh] - (stderr) Error: failed to interpolate file: interpolation error for [DOCKER_APP_TAG] (/var/www/.env:1020): required variable DOCKER_APP_RELEASE is missing a value: error (/var/www/.env:1020)

Could you point out what I am missing here, please?

dansup (approved):

dansup:
cc @jippi

PR #5470, opened 2025-01-14
Fix text no posts archived

Medformatik:
On your own profile page in the “Archives” tab it currently displays “We can't seem to find any posts you have bookmarked”. This is the same text as in the “Bookmarks” tab. I have therefore fixed the text accordingly.


Medformatik:
The `profile.chunk.5a5ea597e5286899.js` file also contains this text, but the file seems to be generated, which is why I have not changed the text there. This is in accordance with the [Contributing Guidelines](https://github.com/pixelfed/pixelfed/blob/staging/CONTRIBUTING.md).

cemreefe (approved):

dansup (approved):

Issue #5474, opened 2025-01-15
Bug: New accounts can have the same email address but a different case


BumbleSusan:
When registering a new user account, it is possible to use the same email address but a different case.

Example:  me@somewhere.com and create a new account as Me@somewhere.com

email addresses should be compared case insensitive? 

v0.12.4 with Curated Onboarding

PR #5491, opened 2025-01-16
check for existing case insensitive email match before signing up

punkMonk123:
Identified by this bug report: https://github.com/pixelfed/pixelfed/issues/5474#issue-2789453855

This change converts the email address to lowercase before checking its existence in the database, ensuring uniqueness of email addresses

### DIFF
--- a/app/Console/Commands/UserCreate.php
+++ b/app/Console/Commands/UserCreate.php
@@ -71,7 +71,7 @@ class UserCreate extends Command
 
         $email = $this->ask('Email');
 
-        if(User::whereEmail($email)->exists()) {
+        if(User::whereRaw('lower(email) = ?', [strtolower($email)])->exists()) {
             $this->error('Email already in use, please try again...');
             exit;
         }


app/Console/Commands/UserCreate.php
```diff
@@ -71,7 +71,7 @@ public function handle()
 
         $email = $this->ask('Email');
 
-        if(User::whereEmail($email)->exists()) {
+        if(User::whereRaw('lower(email) = ?', [strtolower($email)])->exists()) {
```
cemreefe, line 74: I would keep this as it was and have this change on line 72 as per my next comment.
```suggestion
        if(User::whereEmail($email)->exists()) {
```

app/Console/Commands/UserCreate.php
```diff
@@ -71,7 +71,7 @@ public function handle()
 
         $email = $this->ask('Email');
```
cemreefe, line 72: I'd suggest to make the conversion here to keep the separation of functionalities cleaner. We never need a case-sensitive email.

```suggestion
        $email = strtolower(trim($this->ask('Email')));
```

jbergler:
I wanted to report a related issue (login is case sensitive for the email field) which is how I found this PR.

Should this change include fixes for existing non-lower case emails in the database or does that warrant a separate issue / PR?

intentionally-left-nil:
There's a lot more to normalizing an email address than just lowercasing the whole thing. If we're going to address this issue, I'd prefer to see a robust implementation, accounting for edge cases. For example, how does toLower work here with various languages? Do you e.g. want to support punycode and normalize that? What about 0+200B (zero width whitespace) etc etc

app/Console/Commands/UserCreate.php
intentionally-left-nil: An email address absolutely can be case sensitive. Only the domain must be case-insensitive

dansup (changes requested):
Target the staging branch

dansup:
This is only for the user:create command, which the admin has full control over, I think you may have wanted to change the RegisterController or other places we have this logic.

Issue #5236, opened 2024-07-21
Instagram import not working: Only five posts shown for review out of hundred in Zip file


hkuril:
I’ve created a new account on pixelfed.social and I’m trying to import my Instagram posts. I followed the instructions (https://pixelfed.social/site/kb/import). I have a JSON file and by inspecting the file I can see it has my full post history (maybe ~100 posts). But the import screen only shows five posts 
<img width="626" alt="Pixelfed.social import screen" src="https://github.com/user-attachments/assets/b878c72f-2830-487c-aaa8-5eb989871598">.

I can provide the JSON file but would rather not post it here—it has everything including messages, likes, etc.

arnaud-jacquemin:
Hello,

I think I've found the reason why only a minority of posts are taken into account by the importer. It filters the pictures and videos based on their file extension, and only .png, .jpg ou .mp4 are imported. 
It is hardcoded here :
https://github.com/pixelfed/pixelfed/blob/25d065286c7b9844776b70b13588ba9e843d2b7e/resources/assets/components/AccountImport.vue#L375

But most of the files in my Instagram exports have .webp extension, so they are ignored by Pixelfed.

hkuril:
Aha, this seems correct! This should be an easy fix, no? Because Pixelfed already supports webp.

unsolaci:
I've hit a similar issue, so I thought I'd just convert the `.webp` files in the Instagram export.

But `dwebp` would complain about the file type, which lead me to run `file` on them, and to my surprise, those `.webp` files were in fact...**JPEG**:

```
$ file \[redacted\].webp
[redacted].webp: JPEG image data, JFIF standard 1.01, aspect ratio, density 1x1, segment length 16, progressive, precision 8, 1440x1079, components 3
$ exiftool \[redacted\].webp | grep -i 'type'
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
```

Not sure if this is the case for others as well, but if so, I wouldn't be surprised if the Pixelfed importer got confused by it, especially when it seems to be based on file extensions.

hkuril:
This was also the same for me.

[...]

PR #5510, opened 2025-01-18
[Improvement] Add note about import format limitations

codeinabox:
In trying to import photos from Instagram, I came across this issue #5236. This adds a note to the help centre to reflect this limitation.

### DIFF
--- a/resources/views/site/help/import.blade.php
+++ b/resources/views/site/help/import.blade.php
@@ -25,7 +25,7 @@
             <span>Navigate to the <a href="/settings/import" class="font-weight-bold">Import</a> settings page</span>
         </li>
         <li class="">
-            <span>Follow the instructions and import your posts 🥳</span>
+            <span>Follow the instructions and import your posts 🥳 NB: The export may contain files in the webp or heic file formats, this media is not supported by the import</span>
         </li>
     </ol>
     <hr class="mb-4" />


peter76:
Is it maybe a good idea to add instructions as well how to fix this from the command line?

codeinabox:
> Is it maybe a good idea to add instructions as well how to fix this from the command line?

Good point. What are the instructions to manually fix the file?

peter76:
> > Is it maybe a good idea to add instructions as well how to fix this from the command line?
> 
> Good point. What are the instructions to manually fix the file?

This worked for me:

- Unzip your downloaded Instagram archive

- Run the following command in the media directory to rename all .webp files to .jpg

`find . -name "*.webp" -exec rename -v .webp .jpg {} \;`

- Run this in the your-instagram-activity/content folder:

`sed -i 's/.webp/.jpg/g' posts_1.json`

- On linux, from within the unzipped folder, select all folders, right click and create archive.
  Import this archive in PixelFed. If you select the unzipped folder and zip that, the directory  structure is not the same

codeinabox:
@peter76 I tried that on Mac OS but on uploading the file I get the error "The .zip archive you uploaded is corrupted, or is invalid. We cannot process your import at this time". Also, does this work for heic files?

peter76:
@codeinabox How did you rezip it on macos? I'll check on my wife's laptop what is going on. I'm not sure if this works with heic filkes, it works with webp as these ar just jpg with a wrong extension, can you check what

`file nameofpicture.heic`

outputs?

codeinabox:
> @codeinabox How did you rezip it on macos? I'll check on my wife's laptop what is going on. I'm not sure if this works with heic filkes, it works with webp as these ar just jpg with a wrong extension, can you check what
> 
> `file nameofpicture.heic`
> 
> outputs?

The heic files output the following:
> JPEG image data, JFIF standard 1.01, aspect ratio, density 1x1, segment length 16, progressive, precision 8, 1440x1799, components 3

As for compressing it, I discovered the issue was the directory structure.

peter76:
> > @codeinabox How did you rezip it on macos? I'll check on my wife's laptop what is going on. I'm not sure if this works with heic filkes, it works with webp as these ar just jpg with a wrong extension, can you check what
> > `file nameofpicture.heic`
> > outputs?
> 
> The heic files output the following:
> 
> > JPEG image data, JFIF standard 1.01, aspect ratio, density 1x1, segment length 16, progressive, precision 8, 1440x1799, components 3
> 
> As for compressing it, I discovered the issue was the directory structure.

@codeinabox Good you solved the compressing issue! As for the output of the heic files, this means that they are just wrongly named jpg files, just as the webp files. If you replace .webp with .heic in my two original commands they work for the .heic files:

`find . -name "*.heic" -exec rename -v .heic .jpg {} \;`
`sed -i 's/.heic/.jpg/g' posts_1.json`

Regards, Peter

codeinabox:
Thinking about this more, I am inclined to close this as instead of documenting the limitations, I'd argue it would make more sense to add logic in the import code for mapping heic and webp files to jpg. Thoughts @dansup?

villasv:
> As for compressing it, I discovered the issue was the directory structure.

@codeinabox I believe I'm going through the same issue. What was the fix?

EDIT: well, embarrassingly the easiest way to fix it was to download the IG export using Firefox instead of Safari (because Safari unzips automatically)

idanoo (approved):

peter76:
> Thinking about this more, I am inclined to close this as instead of documenting the limitations, I'd argue it would make more sense to add logic in the import code for mapping heic and webp files to jpg. Thoughts @dansup?

@codeinabox at least maybe rephrase your commit, it is not lack of support of the import, but just wrong file extensions, which actually is a bug in the Instagram export, but guess they will not bother:-)

Regards, Peter

shleeable:
As raised, this PR might not be committed as is.. I think `Certain media types might not be supported by the import.`

Might be worth just creating a ticket to support multiple formats and fix it at the source.

dansup:
We now support these formats.

PR #5582, opened 2025-01-24
fix failed import when caption is null

PhieF:
I'm trying to import my instagram posts but running

```
php artisan app:transform-imports
```
will fail because caption cannot be null

this should do the trick

this bug was happening on current stable pixelfed, not sure it is happening in dev

### DIFF
--- a/app/Console/Commands/TransformImports.php
+++ b/app/Console/Commands/TransformImports.php
@@ -103,7 +103,7 @@ class TransformImports extends Command
                 continue;
             }
 
-            $caption = $ip->caption != null ? $ip->caption : "";
+            $caption = !is_null($ip->caption) ? $ip->caption : "";
             $status = new Status;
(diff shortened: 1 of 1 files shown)


app/Console/Commands/TransformImports.php
```diff
@@ -103,7 +103,7 @@ public function handle()
                 continue;
             }
 
-            $caption = $ip->caption;
+            $caption = $ip->caption != null ? $ip->caption : "";
```
idanoo: I would consider doing a stricter null check 

```suggestion
            $caption = $ip->caption ?? "";
```

Or same thing but more explicit

```suggestion
            $caption = !is_null($ip->caption) ? $ip->caption : "";
```

dansup (changes requested):
target the staging branch

PR #5608, opened 2025-01-26
[Improvement] Generic OIDC Support

halkeye:
Generic OIDC support (My usecase is authentik). Assumes oidc provider does everything with post. If its a oidc provider that doesn't, then league has plugins you can install to switch. I didn't handle those scenarios though. I chose league/oauth2-client as it seems other league dependencies are used
https://laravel.com/docs/11.x/socialite might be a better direction, but it seems harder to configure

I don't have any previous laravel experience, so even more open to feedback

### DIFF
--- a/app/Http/Controllers/RemoteAuthController.php
+++ b/app/Http/Controllers/RemoteAuthController.php
@@ -14,6 +14,7 @@ use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Str;
+use App\Rules\PixelfedUsername;
 use InvalidArgumentException;
 use Purify;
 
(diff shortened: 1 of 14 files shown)


osresearch:
This looks far less intrusive than my OIDC patch in https://github.com/pixelfed/pixelfed/pull/3436 (which is now two thousand commits behind main).  I'm also not at all versed with PHP or Laravel.

The banned email checks seems like something that the realm admin would handle, rather than testing in pixelfed?

An additional feature that was in my patch was a `roles` claim that allowed the admin status to be controlled in the OIDC realm as well.  Although a commenter mentioned that maybe the way it was being done was specific to keycloak, so I'm not sure if there is a better way. https://github.com/pixelfed/pixelfed/pull/3436/files#r1414138544

halkeye:
> well. Although a commenter mentioned that maybe the way it was being done was specific to keycloak, so I'm not sure if there is a better way. https://github.com/pixelfed/pixelfed/pull/3436/files#r1414138544

roles/group is always interesting as its never the same between apps. I personally like a "groups" scope, and that returns a groups array in the user info. But its not at all consistent, and for now, i'm thinking i want to keep the PR basic and can be expanded on later.

> The banned email checks seems like something that the realm admin would handle, rather than testing in pixelfed?

I mean yea, I'd leave that up to oidc, but i saw it in the remote mastodon thing, which isn't admin'd the same way, so probably not needed.

halkeye:
I honestly think its pretty close. I have to make sure i have a test case for both new user and existing user. I think I have new user but i'm not certain db gets reset to be 100% sure yet.

Y0ngg4n:
@halkeye Can you slightly estimate how long it will take to add OIDC support to pixelfed? Don´t want to be pushy, there is just different information on different issues.

halkeye:
No?
I have a few more tests I want to write but it'll depend on what the main authors say and so far I haven't heard anything

halkeye:
okay @dansup, I think its ready to go now

fibcom-bp:
@halkeye do you think the PF_OIDC_ envs should be added to the env example files like https://github.com/pixelfed/pixelfed/blob/dev/.env.docker and other docs?

halkeye:
Maybe? I didn't even know that file existed. I'd like to see it working before advertising and having people use it.

bil-ash:
@dansup @halkeye Any chance of this getting merged soon?

halkeye:
🤷 i don't really have any say in this. I see composer needs updating. I can try to update it when needed.

arnaud-jacquemin:
Would some testing help with the merge process? Please let us know. We (Raoull) have a test instance, an identity provider, and we're willing to try this branch if that can help bring OIDC support to Pixelfed.
Thanks in advance!

halkeye:
I'll try update compose files in the next few days. I'm tempted to deploy it anyways to my own instance.

halkeye:
https://github.com/halkeye/docker-pixelfed/pkgs/container/docker-pixelfed/402429241?tag=add-generic-oidc.1

i have a docker image up....

okay its tricker to build than i expected

halkeye:
okay, ghcr.io/halkeye/docker-pixelfed:add-generic-oidc.4 works in apache mode

dansup (approved):
Amazing work, thank you ❤️

PR #5621, opened 2025-01-27
Minor code reviews #1

shleeable:
See comments


shleeable:
I believe these should all be good to merge.

PR #5626, opened 2025-01-27
Update French translation

SebouChu:
Update French translations based on files on this day!


dansup:
Use https://translate.pixelfed.org

Issue #5601, opened 2025-01-26
Mail Settings fail to load


cadaniel:
I'm trying to get SMTP working, and it seems to not be loading the variables correctly.

In my .env
```
# Mail Configuration (Post-Installer)
MAIL_DRIVER=smtp
MAIL_HOST=smtp.DOMAIN.TLD
MAIL_PORT=465
MAIL_ENCRYPTION=ssl
MAIL_USERNAME="EMAIL"
MAIL_PASSWORD="PASSWORD"
MAIL_FROM_ADDRESS="EMAIL"
MAIL_FROM_NAME="Pixelfed - House of Nuts"

```

however, when I go to look at the diagnostics page to view the variables

```
MAIL 	MAIL_DRIVER 	""
```
Same for mail host and port. Oddly,  from address and from name appear correctly and update when I make changes in the .env file

PhieF:
I'm also unable to load mail settings, seems that changed in .env are not taken into account even after a 

`php artisan config:cache`

cadaniel:
I'm noticing the same behaviour for OPEN_REGISTRATION now

PhieF:
@cadaniel are you using docker ?

cadaniel:
@PhieF No, I'm running locally on a VPS. Checked out the latest version per the documentation

PhieF:
OK I've found the issue:

the diagnostic page doesn't display the info well, I've sent a pr for that

however for you to find your issue:

each time you modify .env, run this
`php artisan config:cache && php artisan cache:clear`

if you need to display more log:

set `APP_DEBUG="true"` in your .env, it will display more info in error pages
then once again

`php artisan config:cache && php artisan cache:clear`

be also careful to run the latest version of pixelfed `0.22.4`

@cadaniel

cadaniel:
> OK I've found the issue:
> 
> the diagnostic page doesn't display the info well, I've sent a pr for that
> 
> however for you to find your issue:
> 
> each time you modify .env, run this `php artisan config:cache && php artisan cache:clear`
> 
> if you need to display more log:
> 
> set `APP_DEBUG="true"` in your .env, it will display more info in error pages then once again
> 
> `php artisan config:cache && php artisan cache:clear`
> 
> be also careful to run the latest version of pixelfed `0.22.4`
> 
> [@cadaniel](https://github.com/cadaniel)

I'm seeing the same behaviour and nothing in the logs.

[...]

PR #5632, opened 2025-01-28
fix diagnostic page for mail conf

PhieF:
it doesn't fix mail sending but only the diagnostic page

https://github.com/pixelfed/pixelfed/issues/5601


shleeable:
LGTM

dansup (changes requested):
target the staging branch

PR #5651, opened 2025-01-31
App Register API

dansup:
Add new app register API to allow in-app registration.

To keep the user in the app during the flow as much as possible, while allowing for captcha support, we open a browser session on the pixelfed server to enter their email address and handle the captcha if required.

They are redirected back to the app using a Deep Link and their email as a query parameter, while the backend emails them a 6 digit code they input into the app to verify their session.

From their they can set a username, password and additional optional onboarding steps.

Todo:

- [x] Final registration
- [x] Optional onboarding steps (avatar, bio)
- [x] Open registration detection in app

### DIFF
--- a/app/Http/Controllers/AppRegisterController.php
+++ b/app/Http/Controllers/AppRegisterController.php
@@ -2,11 +2,15 @@
 
 namespace App\Http\Controllers;
 
-use Illuminate\Http\Request;
-use App\Models\AppRegister;
 use App\Mail\InAppRegisterEmailVerify;
+use App\Models\AppRegister;
+use App\User;
+use App\Util\Lexer\RestrictedNames;
+use Illuminate\Http\Request;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Facades\Mail;
+use Purify;
 
 class AppRegisterController extends Controller
 {
@@ -15,9 +19,10 @@ class AppRegisterController extends Controller
         abort_unless(config('auth.iar') == true, 404);
         // $open = (bool) config_cache('pixelfed.open_registration');
         // if(!$open || $request->user()) {
-        if($request->user()) {
+        if ($request->user()) {
             return redirect('/');
         }
+
         return view('auth.iar');
     }
 
@@ -40,11 +45,12 @@ class AppRegisterController extends Controller
 
         $exists = AppRegister::whereEmail($email)->where('created_at', '>', now()->subHours(24))->count();
 
-        if($exists && $exists > 3) {
+        if ($exists && $exists > 3) {
             $errorParams = http_build_query([
                 'status' => 'error',
-                'message' => 'Too many attempts, please try again later.'
+                'message' => 'Too many attempts, please try again later.',
             ]);
+
             return redirect()->away("pixelfed://verifyEmail?{$errorParams}");
         }
 
@@ -53,7 +59,7 @@ class AppRegisterController extends Controller
         $registration = AppRegister::create([
             'email' => $email,
             'verify_code' => $code,
-            'email_delivered_at' => now()
+            'email_delivered_at' => now(),
         ]);
 
         try {
@@ -62,8 +68,9 @@ class AppRegisterController extends Controller
             DB::rollBack();
             $errorParams = http_build_query([
                 'status' => 'error',
-                'message' => 'Failed to send verification code'
+                'message' => 'Failed to send verification code',
             ]);
+
             return redirect()->away("pixelfed://verifyEmail?{$errorParams}");
         }
 
@@ -72,7 +79,7 @@ class AppRegisterController extends Controller
         $queryParams = http_build_query([
             'email' => $request->email,
             'expires_in' => 3600,
-            'status' => 'success'
+            'status' => 'success',
         ]);
 
         return redirect()->away("pixelfed://verifyEmail?{$queryParams}");
@@ -80,9 +87,11 @@ class AppRegisterController extends Controller
 
     public function verifyCode(Request $request)
     {
+        abort_unless(config('auth.iar') == true, 404);
+
         $this->validate($request, [
             'email' => 'required|email:rfc,dns,spoof,strict|unique:users,email',
-            'verify_code' => ['required', 'digits:6', 'numeric']
+            'verify_code' => ['required', 'digits:6', 'numeric'],
         ]);
 
         $email = $request->input('email');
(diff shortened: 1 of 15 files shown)


config/auth.php
```diff
@@ -112,4 +112,5 @@
         ],
     ],
 
+    'iar' => env('APP_REGISTER', false),
```
jippi: For ease of understanding the code base an control flows, I would love for us to be a bit more verbose with our configuration keys, it will read more fluently and avoid having to maintain a map of short names to actual meaning

```suggestion
    'in_app_registration' => env('APP_REGISTER', false),
```

jippi: Is there a fixed convention between `PF_`, `APP_` and no prefix in env keys? 

Perhaps it would be worth naming the PHP and env settings consistently, making it easier to map the linage of the various layers of configuration as well

```suggestion
    'in_app_registration' => env('IN_APP_REGISTRATION', false),
```

app/Http/Controllers/AppRegisterController.php
```diff
@@ -0,0 +1,100 @@
...
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Mail;
+
+class AppRegisterController extends Controller
+{
+    public function index(Request $request)
+    {
+        abort_unless(config('auth.iar') == true, 404);
+        // $open = (bool) config_cache('pixelfed.open_registration');
+        // if(!$open || $request->user()) {
```
jippi: nit: remove dead code if its not going to be used

config/auth.php
jippi: Do we need to cast to `bool` here to ensure the ENV setting (a string) is correctly evaluated down to a type-safe boolean in code?

app/Http/Controllers/AppRegisterController.php
```diff
@@ -0,0 +1,100 @@
...
+use App\Models\AppRegister;
+use App\Mail\InAppRegisterEmailVerify;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Mail;
+
+class AppRegisterController extends Controller
+{
+    public function index(Request $request)
+    {
+        abort_unless(config('auth.iar') == true, 404);
```
jippi: (See comment in config/auth.php first)

Could we simplify this like below if we type-cast the env key/config to boolean - it would return more clearly / naturally imo

```suggestion
        abort_unless(config('auth.iar'), 404);
```

app/Http/Controllers/AppRegisterController.php
```diff
@@ -0,0 +1,100 @@
...
+        if ((bool) config_cache('captcha.enabled') && (bool) config_cache('captcha.active.register')) {
+            $rules['h-captcha-response'] = 'required|captcha';
+        }
+
+        $this->validate($request, $rules);
+
+        $email = $request->input('email');
+        $code = str_pad(random_int(0, 999999), 6, '0', STR_PAD_LEFT);
+
+        $exists = AppRegister::whereEmail($email)->where('created_at', '>', now()->subHours(24))->count();
```
jippi, line 54: should we ensure the email is case in-sensitive to avoid bad actors potentially using `DaNSuP@pixelfed.social` and other case iterations? Postgres in particular is case sensitive on strings unless configured otherwise

app/Http/Controllers/AppRegisterController.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if($exists && $exists > 3) {
+            $errorParams = http_build_query([
+                'status' => 'error',
+                'message' => 'Too many attempts, please try again later.'
+            ]);
+            return redirect()->away("pixelfed://verifyEmail?{$errorParams}");
+        }
+
+        DB::beginTransaction();
```
jippi: I think this would be better put before the `$exists` check so the check + insert is done together in a single transaction to avoid any race conditions

app/Http/Controllers/AppRegisterController.php
```diff
@@ -0,0 +1,100 @@
...
+        $email = $request->input('email');
+        $code = $request->input('verify_code');
+
+        $exists = AppRegister::whereEmail($email)
+            ->whereVerifyCode($code)
+            ->where('created_at', '>', now()->subMinutes(60))
+            ->exists();
+
+        return response()->json([
+            'status' => $exists ? 'success' : 'error',
```
jippi, line 116: From a UX point of view, should we have 3 outcomes her perhaps? we would be able to provide better feedback to users like that

the code+email combo does NOT exist (error)
the code+email combo does exist, but too old (expired)
the code+email combo does exist, and NOT expired (success)

database/migrations/2025_01_28_102016_create_app_registers_table.php
```diff
@@ -0,0 +1,32 @@
...
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('app_registers', function (Blueprint $table) {
+            $table->id();
+            $table->string('email');
+            $table->string('verify_code');
+            $table->unique(['email', 'verify_code']);
+            $table->timestamp('email_delivered_at')->nullable();
```
jippi, line 19: is this _technically_ not the sent time, rather than delivered? 

```suggestion
            $table->timestamp('email_sent_at')->nullable();
```

app/Mail/InAppRegisterEmailVerify.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->code = $code;
+    }
+
+    /**
+     * Get the message envelope.
+     */
+    public function envelope(): Envelope
+    {
+        return new Envelope(
+            subject: config('pixelfed.domain.app') . ' - Verify Your Email Address',
```
jippi: Should these subjects be localizable (so a danish site can have danish email subjects)?

jippi (commented):
LMK what you think, overall it looks pretty good!

app/Http/Controllers/AppRegisterController.php
dansup: This isn't dead, just temporary while I test on the closed staging server.

app/Http/Controllers/AppRegisterController.php
```diff
@@ -159,9 +162,32 @@ public function onboarding(Request $request)
 
         sleep(10);
```
jippi: Whats the purpose of this sleep?

dansup: After the User is created, the UserObserver generates the public/private key, and associated models. Normally we don't need this, but since we are immediately generating an authToken and calling verify_credentials in the app.

PR #5681, opened 2025-02-05
[Translation] Make appearance settings translateable

maco:
The "UI Settings" menu item is currently hardcoded, not translateable, so this fixes that and adds the Spanish translation.

I also changed it from "UI Settings" to "Appearance" because I think it'll be better for non-technical users.

However, I haven't figured out how to test this locally because I don't understand how the translation js files are generated.

### DIFF
--- a/resources/assets/components/partials/navbar.vue
+++ b/resources/assets/components/partials/navbar.vue
@@ -270,7 +270,7 @@
 										<li class="nav-item">
 											<a class="nav-link" href="/i/web" @click.prevent="openUserInterfaceSettings">
 												<span class="icon text-lighter"><i class="far fa-brush"></i></span>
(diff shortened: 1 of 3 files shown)


maco:
@Uthanien by the way, mind reviewing this? Might as well review each other's translations, right?

Uthanien:
> @Uthanien by the way, mind reviewing this? Might as well review each other's translations, right?

No problem, I'll add these new variables

maco:
I found out how to rebuild the front end and found I have a bug in the .vue file, and either I still need to figure out a little more about that build system, or I'm using translatable strings wrong in the modal. So I'll poke at it more later tonight or tomorrow.

dansup (approved):

PR #5685, opened 2025-02-05
laravel/pint Github Action

shleeable:
Require installing Pint `composer global require laravel/pint`

and then running against staging.. this will likely break all of the existing PRs but would be good to be done sooner than later


jippi:
Other PRs can cherry-pick this into their branch, and then run the linter locally to update their source code - should minimize conflicts greatly :)

PR #5722, opened 2025-02-10
Fix curl

dansup:


bagder:
You want CURLOPT_SSL_VERIFYHOST set TRUE as well!

PR #5741, opened 2025-02-15
[Bug Fix] ProfileController: extend Status types visible in the ATOM feed.

### LINKED ISSUE #5695: RSS feed does not include video posts
Example: [this post](https://pxlfd.ca/i/web/post/792588034524958505) is missing from [the profile feed](https://pxlfd.ca/users/victor.atom)

rm-yakovenko:
Fixes #5695.

### DIFF
--- a/resources/views/atom/user.blade.php
+++ b/resources/views/atom/user.blade.php
@@ -29,7 +29,7 @@
                     <![CDATA[
                     <div class="media-gallery">
                         @foreach($item['media_attachments'] as $media)
-                            <img class="media-item" src="{{ $media['url'] }}" alt="{{ $media['description'] ?? '' }}">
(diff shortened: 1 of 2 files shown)


resources/views/atom/user.blade.php
```diff
@@ -43,7 +43,6 @@
                     <media:content
                         url="{{ $media['url'] }}"
                         type="{{ $media['mime'] }}"
-                        medium="image"
```
rm-yakovenko, line 46: Since not only images can land here, let the feed reader decide how to display media content.

dansup (changes requested):
The `preview_url` can contain a placeholder image, and doesn't properly support what you are trying to accomplish.

dansup:
When you do fix this, please ping me and I will re-open this.

PR #5765, opened 2025-02-20
[Bug Fix] Give execute permissions to folders

### LINKED ISSUE #4275: [Web Bug]: Picture display broken after upgrade to v0.11.5 (folder rights issue)
Hello,

On Yunohost installations, since version 0.11.5 new uploaded pictures are broken. They are not shown, neither the preview is.  Even in compose view.
Alternative text is working.
I don't really know how to debug this. No horizon failed job, no supervisor issue, I can't find laravel log, no web console error…
No other error to report.
You can see an example here : https://photo.lapineige.fr/i/web/post/548580903387685397

Is that a know Pixelfed bug ?
How can I diagnose this ?

Thank you

---

Pixelfed : 0.11.5
Postgres: 13.9
PHP: 8.1 - recently upgraded, before Pixelfed 0.11.5 we were using php 8.0
Laravel : 9.52.4

lullis:
nginx (and presumably every other web server) returns 403 for every url on app/public/storage because of the strict permissions. 

Allowing folders to be created with the execute bit set is enough to solve this without making it possible to list its contents.

Fixes #4275

### DIFF
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -53,7 +53,7 @@ return [
                 ],
                 'dir' => [
                     'public' => 0755,
-                    'private' => 0700,
+                    'private' => 0711,
                 ],
             ],
         ],


shleeable:
This issue catches a lot of people (including myself). This bug fix is perfectly fine to me because my understanding is this is always temporary files...

dansup (approved):

PR #5829, opened 2025-03-06
[Translation] transalte notifications

eufelipemateus:
I recently translated the notifications on my instance and would like to share this improvement with the community. I'm open to suggestions and feedback to further refine the translation!


dansup:
Please do not edit this file, it is generated from the `resources/lang/*` files. You need to edit the `resources/lang/pt/*.php` files.

PR #5837, opened 2025-03-07
[Translation] translate apparence pt

eufelipemateus:
This PR translates the system's language appearance variables into Portuguese, ensuring a better experience for Portuguese-speaking users.


dansup:
Please do not edit this file, it is generated from the `resources/lang/*` files. You need to edit the `resources/lang/pt/*.php` files.

PR #5840, opened 2025-03-08
[Improvement] Replicate mastodon list

eufelipemateus:
## Description
This PR improves how the followers and following lists are returned in ActivityPub, adopting the same behavior as Mastodon. Now, the API correctly displays the users a profile follows and those who follow it, enhancing integration and user discovery.

To ensure privacy, if a user chooses to hide this information, the API will return an empty list, respecting their settings.

Examples of the new listing are available in the screenshots below.

## Screenshots
image

Captura de tela de 2025-03-08 12-26-35


dansup (changes requested):
This doesn't properly handle authorizedFetch, which is super important.

Issue #5810, opened 2025-02-27
[Bug] "Shared by 2 others" not clickable on profile tab 'Likes'


GiantCrocodile:
Hello,

when you go to your profile in the web ui and click on the tab 'Likes', you can't press on the "Shared by 2 others" link on some picture you have liked. Nothing happens after clicking on it.

![Image](https://github.com/user-attachments/assets/8225a353-dc4f-4306-b9e1-db8c02033efa)

Happens with v0.12.4.

Best Regards,

shleeable:
Please reopen with updated info if still relevant to the latest release

PR #5879, opened 2025-03-19
[Improvement] Fix for issue #5810: Bug in the user profile likes list display and interaction + reblog display adjustment

eufelipemateus:
# PR Description  
### Identified Problem  
This PR fixes issue #5810, which caused inconsistencies in the display and interaction of the likes list on the user profile. The bug occurred because:  

- The likes list in the user profile was not correctly using the `feed` variable. As a result, posts were listed correctly, but interactions (like, reblog, etc.) did not work properly.  
- When a user tried to like or view the list of shared posts, the reference to the post was incorrect or not found, especially when the user's post feed was loaded.  
- Additionally, the reblog display did not properly differentiate when the user viewing the post was also the one who had reblogged it.  

### Implemented Solutions  
✅ Adjusted the logic of the user profile likes list to correctly use the `feed` variable, ensuring proper display and interaction synchronization.  
✅ Fixed the reference to posts in the likes list to prevent errors when liking, reblogging, or viewing shared posts.  
✅ Updated the backend to correctly return the `reblogged` value when the user who shared the post is the same as the one viewing it. Instead of simply displaying `1`, it will now show `me` when the viewer is the one who reblogged the post.  

With these fixes, users will be able to interact properly with the likes list in their profile, and the reblog display will be more intuitive by indicating when the user is responsible for the reblog.


shleeable:
@eufelipemateus do you use discord? please join ours. Thank you for your PRs.

eufelipemateus:
> @eufelipemateus do you use discord? please join ours. Thank you for your PRs.
I appreciate the invitation, but I don't use Discord. Also, I don't know speak English. 😊

dansup:
Properly fixed in [5ddb6d842](https://github.com/pixelfed/pixelfed/commit/a25afa0442c6042a951bca5daec2cc912de8284f)

PR #5912, opened 2025-04-02
[Bug Fix] fix query pinned posts

eufelipemateus:
This PR fixes a bug that prevented multiple posts from being pinned. Query was wrong!

### DIFF
--- a/app/Services/StatusService.php
+++ b/app/Services/StatusService.php
@@ -203,8 +203,7 @@ class StatusService
 
     public static function isPinned($id)
     {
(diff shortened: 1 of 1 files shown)


dansup:
Fixed in https://github.com/pixelfed/pixelfed/commit/2f655d000844f4728c40407adc1239a5b5912b9f

There were a few other improvements I needed to make to make this work properly, for example, including making pinned posts work with pagination, and recalculating the pinned_order when one is unpinned.

You did a great job on the foundation for this feature, thanks a lot <3

Issue #5070, opened 2024-05-04
Muting a user should hide old notifications from that user


steinarb:
Replies/comments from a muted user still shows up in pixelfed notifications.

I don't think they should. Muted is muted.

PR #5918, opened 2025-04-05
[Improvement] Remove old notifications

eufelipemateus:
This PR try fix this issue: #5070


dansup (changes requested):
We need a better implementation, this would fail on larger servers with millions of notifications like Pixelfed.social.

We should dispatch a job to filter this with a more efficient query.

Issue #5309, opened 2024-09-26
[Server Bug]: Can't enable AVIF media file (Pixelfed 0.12.3~ynh2)


jothgard:
Hello, are running Pixelfed 0.12.3 on Yunohost!

Seems that i cannot activate the media file support for AVIF in WebAdmin panel.

Have added following line in /var/www/pixelfed/.env

MEDIA_TYPES=image/jpeg,image/avif,image/png,image/gif,image/webp,video/mp4,video/mov

But seems still not be activating avif support.

What are im missing here?

Using Pixelfed 0.12.3~ynh2

dansup:
Hi, sorry about this issue you are running into.

I'm in the process of shipping AVIF support using libvips, and expect it to ship later this weekend!

I'll keep this issue open, and let you know when AVIF support is fixed and ready for production.

jothgard:
> Hi, sorry about this issue you are running into.
> 
> I'm in the process of shipping AVIF support using libvips, and expect it to ship later this weekend!
> 
> I'll keep this issue open, and let you know when AVIF support is fixed and ready for production.

Oh okey, thanks for a fast reply! GREAT work on Pixelfed! And so happy to host it with just one click install on YH :)!

ning1no:
Hi, I am having the same issue with MOV & HEIC, is it the same thing that AVIF?

jothgard:
> Hi, I am having the same issue with MOV & HEIC, is it the same thing that AVIF?

I think so!

rexbron:
Still an issue on pixelfed.social as of December 2025

PR #5934, opened 2025-04-11
[Bug Fix] Fix unchecked media types HEIC and MOV in admin settings

Niols:
Hello there!

Thank you very much for this wonderful piece of software. Here is my very first contribution; hopefully it is useful!

---

Currently, the admin settings view of medias mime types seems to ignore HEIC and MOV files, even though I have enabled then via `.env` variables. I believe this is “only” a front-end problem, in the sense that the front-end will show HEIC and MOV as unchecked no matter what, even when they are enabled in the back-end. Moreover, I think that saving the form does work fine for HEIC, but will ignore MOV. However, a subsequent save of the form (where HEIC is disabled by default) might disable HEIC by mistake.

image
_(Image taken from [this Reddit post](https://www.reddit.com/r/PixelFed/comments/1iezxfy/i_cant_enable_avif_heic_or_mov_uploads_the/) that seems to report the same issue.)_

This is possibly what @ning1no is encountering in https://github.com/pixelfed/pixelfed/issues/5309#issuecomment-2455571354, but it is hard to say without more information on what “[it] seems still not be activating avif support” and “the same issue” mean there.

---

The first part of this happens in `fetchInitialData`:

https://github.com/pixelfed/pixelfed/blob/e5dc95953a9574b8cf74a4c08f8d0cbdc393f19d/resources/assets/components/admin/AdminSettings.vue#L1072-L1092

On my instance, querying `/i/admin/api/settings/fetch` by hand does report the right string of mime types, containing `image/heic` and `video/mov`. However, `fetchInitialData` calls `setMediaTypes`:

https://github.com/pixelfed/pixelfed/blob/e5dc95953a9574b8cf74a4c08f8d0cbdc393f19d/resources/assets/components/admin/AdminSettings.vue#L1094-L1104

which will only ever recognise the hard-coded list of types. The current PR simply adds `heic` and `mov` to that list. Another solution would be to grab the keys of the dictionary `mediaTypes` in that same file:

https://github.com/pixelfed/pixelfed/blob/e5dc95953a9574b8cf74a4c08f8d0cbdc393f19d/resources/assets/components/admin/AdminSettings.vue#L906-L915

to limit duplication and avoid re-encountering the same issue in the future when more media types are added. I will happily amend my PR with such a change if it is preferred.

---

The second part of this happens in `activeMediaTypes.get`:

https://github.com/pixelfed/pixelfed/blob/e5dc95953a9574b8cf74a4c08f8d0cbdc393f19d/resources/assets/components/admin/AdminSettings.vue#L1004-L1042

that lists everything but MOV. There too, it might be worth trying to find a way to have a unique source of truth and generate this code from there. However, the `mediaTypes` dictionary might not contain enough information for this.

---

Potentially related: I have tried uploading an HEIC file to my instance, where the back-end does report `image/heic` when I query `/i/admin/api/settings/fetch`, and I get

> The file you are trying to add is not a valid mime type. Please upload a ...,image/heic,... only.

I haven't inspected more because it is maybe fixed by this small PR, but if not I will investigate and report on this.

### DIFF
--- a/resources/assets/components/admin/AdminSettings.vue
+++ b/resources/assets/components/admin/AdminSettings.vue
@@ -1021,16 +1021,20 @@
                         res += 'image/webp,'
                     }
 
-                    if(this.mediaTypes.mp4) {
-                        res += 'video/mp4,'
+                    if(this.mediaTypes.avif) {
+                        res += 'image/avif,'
                     }
 
                     if(this.mediaTypes.heic) {
                         res += 'image/heic,'
                     }
 
-                    if(this.mediaTypes.avif) {
-                        res += 'image/avif,'
+                    if(this.mediaTypes.mp4) {
+                        res += 'video/mp4,'
+                    }
+
+                    if(this.mediaTypes.mov) {
+                        res += 'video/mov,'
                     }
 
                     if(res.endsWith(',')) {
@@ -1096,7 +1100,7 @@
                 if(types && types.length) {
                     types.forEach((type) => {
                         let mime = type.split('/')[1];
-                        if(['jpeg', 'png', 'gif', 'webp', 'mp4', 'avif'].includes(mime)) {
+                        if(['jpeg', 'png', 'gif', 'webp', 'avif', 'heic', 'mp4', 'mov'].includes(mime)) {
                             this.mediaTypes[mime] = true;
                         }
                     })


wolfspyre:
ideally this wouldn’t be buried in code. but rather in a global config/settings file someplace, OR in a db table administratively manipulatable via an admin panel…

having a single source of truth, in an intuitive to find/inspect, and programatically consistent place to lookup… rather than being fragmented in multiple places  is likely to be more beneficial in the long run imo

dansup (approved):

PR #6012, opened 2025-05-22
Refactor users label

eufelipemateus:
### Description

This feature  **refactor the user labels system**, enabling the creation of an **open and flexible interface** that allows for adding multiple labels in the future without requiring changes across multiple parts of the codebase.

A new database table `users_labels` has been created with the following fields:

- `name`
- `label`
- `background_color`
- `text_color`

These fields are used to **dynamically render labels** in the user interface through the `getLabelAttribute()` function, which defines the **business logic** for displaying user labels.

By default, the following labels are included:

- `admin`: indicates administrative users.
- `new`: marks newly registered users.
- `mod`: identifies moderators.

However, the system is designed to easily allow **adding new labels** in the future, with each instance of the application able to configure its **own business rules** as needed.

### Priority Rules

The `getLabelAttribute()` function is designed to return the **most important label** according to a defined priority. For example:

- If a user has both the `admin` and `mod` labels, only the **`admin` label** will be displayed, as it is considered more important.
- The priority order should be configured in the function's return logic to ensure that the most relevant label is always displayed first.

### Motivation

- **Eliminate code duplication** when dealing with multiple labels.
- Facilitate **maintenance and future expansion** of the label system.
- Make label display **more flexible and configurable**.
- Establish a **centralized structure** for label display logic.


eufelipemateus:
I'm considering removing the **user_label** model and table, and instead moving this to a configuration file (_config/user_labels.php_). This change is aimed at improving performance. I'd like to hear your suggestions."

PR #6015, opened 2025-05-27
translate - Remove follow 

eufelipemateus:
This PR transalte remove follow in PT - BR


dansup (changes requested):
Why are you removing this?

PR #6053, opened 2025-07-02
[BugFix] Fix error on Appearance menu, so the translated name is shown

joergi:
> If you want to contribute to this repository, please file your pull request against the staging branch.
As wished, the PR is against staging

at the moment the text for Appearance is not shown.
it only shows: `$t('navmenu.appearance')`
image

there are missing curly braces


resources/assets/components/partials/navbar.vue
```diff
@@ -275,7 +275,7 @@
             hide-footer
             centered
             body-class="p-0 ui-menu"
-            title="$t('navmenu.appearance')">
+            title="{{ $t('navmenu.appearance') }}">
```
eufelipemateus:
```suggestion
           :title="$t('navmenu.appearance')">
```

joergi: do you mind explaining this to me @eufelipemateus?
I was doing exactly how it's done for all the other points in the menu.
(I'm not a vue dev, I'm coming from Java/Kotlin world, but worked with handlebars before, so maybe just send me a relevant link I have to read how it's working differently than that)
Thanks

eufelipemateus: Hi! @joergi  No worries at all i  happy to explain! 😊

In Vue, there are three common ways to render or bind values:

**1. HTML attribute (static):**

```html 
title="Static text" 
```
This is just plain HTML — it doesn’t evaluate any expressions.

**2. Template interpolation (rendered content):**

```html 
{{ $t('navmenu.appearance') }}
```
This is used when you want to print the result inside the template — like between tags:


```html 
<span>{{ $t('some.key') }}</span> 
```

3. Dynamic binding (Vue directive):


```html 
:title="$t('navmenu.appearance')"
```
The `:` is shorthand for `v-bind:`. This tells Vue to evaluate the expression and bind it as the value of the attribute. So in this case, it dynamically sets the `title` attribute based on the translation key.

# 

So since you're binding a dynamic value to a prop (the `title` attribute), you should use **`:title`, not `title="{{ }}"`.**

> I understand you're coming from Java/Kotlin and have used Handlebars,  Vue's {{ }} might look similar but it's only for rendering **visible content**, not attributes.

Here’s a helpful link from the Vue docs that explains this in more detail:
🔗 [https://vuejs.org/guide/essentials/template-syntax.html#attribute-bindings](https://vuejs.org/guide/essentials/template-syntax.html#attribute-bindings)

joergi: Thanks for the long explanation. I will read through it the next days!

eufelipemateus (approved):

dansup (approved):

PR #6098, opened 2025-07-31
Fix hashtags usings scripts outside the BMP on MySQL

bpeel:
The default collation for MySQL/MariaDB seems to be `utf8mb4_unicode_ci` and it looks like this doesn’t try to compare characters that are outside of the basic multilingual plane (ie, codepoints >= 0x10000). This means that any hashtags written in scripts outside the BMP end up being equal to any other hashtags outside the BMP that have the same length. For example, if you look at the [𐑖𐑱𐑝𐑾𐑯 hashtag on pixelfed.social](https://pixelfed.social/i/web/hashtag/%F0%90%91%96%F0%90%91%B1%F0%90%91%9D%F0%90%91%BE%F0%90%91%AF), not only do you get posts related to the Shavian alphabet, you also get [𒆳𒆍𒀭𒊏𒆠](https://en.wikipedia.org/wiki/Babylonia) and a bunch of statuses that are using funky characters to make italicised hashtags. As long as they are also 5 letters long they will be considered the same thing.

Also, if I view a status from Pixelfed using one of these hashtags with [Mastodon](https://tech.lgbt/@kneilknits@pixelfed.social/114942738324307911) then the irrelevant cuneiform hashtag gets added to the bottom of the post.

This patch migrates the `hashtags` table to use the `utf8mb4_unicode_520_ci` collation which does seem to compare these characters properly. I think this should make it so that if anyone uses the Shavian hashtag going forward then it will end up creating a new entry in the hashtags table instead of conflating it with the cuneiform hashtag. I guess it won’t fix any existing posts that have already been conflated.

It seems to work fine with the Postgres backend without any changes. I guess it works because in that case Pixelfed is comparing the hashtags with `ilike` and that presumably doesn’t try to do any fancy collation.

### DIFF
--- a/database/migrations/2025_07_31_164635_change_hashtags_collation.php
+++ b/database/migrations/2025_07_31_164635_change_hashtags_collation.php
@@ -1,0 +1,36 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        if (config('database.default') === 'pgsql')
+            return;
+
+        Schema::table('hashtags', function (Blueprint $table) {
+            $table->string('name')->collation('utf8mb4_unicode_520_ci')->change();
+            $table->string('slug')->collation('utf8mb4_unicode_520_ci')->change();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        if (config('database.default') === 'pgsql')
+            return;
+
+        Schema::table('hashtags', function (Blueprint $table) {
+            $table->string('name')->change();
+            $table->string('slug')->change();
+        });
+    }
+};
--- a/database/migrations/2025_07_31_164635_change_hashtags_collation.php
+++ b/database/migrations/2025_07_31_164635_change_hashtags_collation.php
@@ -1,23 +1,36 @@
 <?php
 
 use Illuminate\Database\Migrations\Migration;
-use Illuminate\Database\Schema\Blueprint;
-use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Facades\DB;
 
 return new class extends Migration
 {
+    /**
+     * The collation that correctly differentiates characters outside the
+     * Basic Multilingual Plane (codepoints >= 0x10000). The historic default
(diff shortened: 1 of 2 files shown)


bpeel:
For what it’s worth, even though the patch changes the database to use a collation based on Unicode 5.20, I think this will fix the problem for scripts added in later versions of Unicode too. See this quote from the [MySQL docs](https://dev.mysql.com/doc/mysql-g11n-excerpt/5.7/en/charset-unicode-sets.html):

> For supplementary characters based on UCA versions higher than 4.0.0 (for example, `xxx_unicode_520_ci`), supplementary characters do not necessarily all have the same collating weight. Some have explicit weights from the UCA `allkeys.txt` file. Others have weights calculated from this algorithm:
>
> ```
> aaaa= base +  (code >> 15);
> bbbb= (code & 0x7FFF) | 0x8000;
> ```

I did a quick test with the Brahmi script which was added in Unicode 6.0 and it seems to work:

```
MariaDB [temptest]> set names utf8;
Query OK, 0 rows affected (0,000 sec)

MariaDB [temptest]> use temptest
Database changed
MariaDB [temptest]> create table t (s1 char(4) character set utf8mb4 collate utf8mb4_general_ci);
Query OK, 0 rows affected (0,043 sec)

MariaDB [temptest]> insert into t values (0xf0918080); # Brahmi character U+11000
Query OK, 1 row affected (0,030 sec)

MariaDB [temptest]> insert into t values (0xf0918081); # Brahmi character U+11001
Query OK, 1 row affected (0,007 sec)

MariaDB [temptest]> select count(distinct(s1)) from t;
+---------------------+
| count(distinct(s1)) |
+---------------------+
|                   1 |
+---------------------+
1 row in set (0,000 sec)
```

The query result shows there is only one distinct value in the table which means it is equating the two Brahmi characters.

If we change the table to use the Unicode 5.20 collation then the result is different:

```
MariaDB [temptest]> alter table t modify column s1 char(4) collate 'utf8mb4_unicode_520_ci';
Query OK, 0 rows affected (0,053 sec)
Records: 0  Duplicates: 0  Warnings: 0

MariaDB [temptest]> select count(distinct(s1)) from t;
+---------------------+
| count(distinct(s1)) |
+---------------------+
|                   2 |
+---------------------+
1 row in set (0,000 sec)
```

Now there are two distinct values, so the two characters are considered different even though they come from Unicode 6.0.

shleeable:
This is a MAJOR change and we will need to investigate to confirm this can be completed safely on an existing instance.

Thank you for the effort into this PR.

shleeable:
Refactored and added tests

PR #6184, opened 2025-09-18
Update DDEV PHP, MariaDB and Redis versions

rossbearman:
The current DDEV config isn't compatible with the composer requirements (PHP > 8.2). This bumps the versions of PHP (8.4), MariaDB (11.8) and Redis (7), as well as tweaking the environment variable overrides to use the `mariadb` database connection and the `phpredis` extension, rather than the `predis` library.

### DIFF
--- a/.ddev/addon-metadata/redis/manifest.yaml
+++ b/.ddev/addon-metadata/redis/manifest.yaml
@@ -1,0 +1,38 @@
+name: redis
+repository: ddev/ddev-redis
+version: v2.1.2
+install_date: "2025-09-18T23:53:47+01:00"
+project_files:
+    - docker-compose.redis.yaml
+    - redis/scripts/settings.ddev.redis.php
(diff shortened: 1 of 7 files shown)


Ramoonus:
@rossbearman why redis 7 and valkey 8?

rossbearman:
@Ramoonus Those are just the latest default versions provided by the `ddev-redis` addon that's already being used by Pixelfed. I updated the addon, which now autogenerates the `docker-compose.redis.yaml`, `manifest.yaml` and command files. 

The default is to use Redis 7 as the backend, but it can be changed using the `ddev redis-backend` command, for example `ddev redis-backend redis:8`.  You can also set this in a local `.ddev/.env.redis` file to always override the default.

There's an [open issue](https://github.com/ddev/ddev-redis/issues/51) on the ddev-redis repo about changing the default to Redis 8, but it's not been done yet.

dansup (approved):

PR #6281, opened 2025-11-05
Larastan: Update Installer.php - typo

shleeable:
`is_writable`


dansup:
It's not actually a [typo](https://www.php.net/manual/en/function.is-writeable.php)

PR #6322, opened 2025-11-09
Fix error for DMs without attachment 

grossermensch:
Fix undefined array key error when processing ActivityPub objects without attachments.

### DIFF
--- a/app/Console/Commands/AddUserDomainBlock.php
+++ b/app/Console/Commands/AddUserDomainBlock.php
@@ -34,7 +34,7 @@ class AddUserDomainBlock extends Command
         $domain = text('Enter domain you want to block');
         $domain = strtolower($domain);
         $domain = $this->validateDomain($domain);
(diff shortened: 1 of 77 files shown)


shleeable:
Thank you for fixing this. I had another branch with a different solution...

but this change might be better than mine?

I was considering adding extra checks to the loop

```
            foreach ($activity['attachment'] as $a) {
                if (!isset($a['mediaType'])) {
                    continue;
                }
                if (!isset($a['url'])) {
                    continue;
                }
                $type = $a['mediaType'];
                $url = $a['url'];
```

grossermensch:
The rest of the flow for DMs seems to work fine. As far as I understand it, starting with php8+, if a key is missing it throws an error and doesn’t proceed. So just fixing the case of no key by working with an empty array should be enough. The less code the better ;)

dansup (approved):

PR #6323, opened 2025-11-09
Fixing Horizon Error for DMs

grossermensch:
Fixing the following horizon error for DMs: 
Error: Class "App\Util\ActivityPub\User" not found in /var/www/pixelfed/app/Util/ActivityPub/Inbox.php:545


shleeable:
this is also resolved in #6299 with other missing classes.

PR #6331, opened 2025-11-12
BREAKING - Docker: Replace existing Dockerfile / docker-compose.yml - using ServerSideUp/PHP base image - Mergable

shleeable:
This will break for anybody using the current out-of-the-box pixelfed but it should also be very simple for them to resolve it as the only things that matter is the .env file, storage and bootstrap/cache folder.

Tested in [https://pixelstaging1.shl.ee/](https://pixelstaging1.shl.ee/) and I'm preparing for "production".

1. This removes a lot of manual scripting (notice all of the deleted files) and replaced with out of the box by ServerSideUp/PHP support. An image build with laravel best practices.
2. PHPRedis should be supported out of the box - faster than predis
3. I've added VIP to this Dockerfile so we can easily test changing the image processing pipeline.
4. Added health checks to all containers.

**This requires matching update to the documentation.**


shleeable:
Video demo

https://github.com/user-attachments/assets/5f4db2f1-2095-4ba4-ae91-fb60065f6414

dansup (approved):

PR #6367, opened 2025-11-21
Enhancement: Create app/Util/storage/safeDeleteDirectory.php with deleteRecursive functions

shleeable:
Context: I've noticed that Fastly Object storage doesn't delete a folder recursively. 
This causes "Storage::deleteDirectory" to fail on fastly, but work as expected on wasabi.

This is a platform agnostic function to delete all of the files/folders on local disk or s3.

Tested on staging environment.
* AWS tested
* Wasabi tested
* fastly tested

### DIFF
--- a/app/Util/storage/safeDeleteDirectory.php
+++ b/app/Util/storage/safeDeleteDirectory.php
@@ -36,6 +36,7 @@ class SafeDeleteDirectory
 
     /**
      * Recursively delete all files and subdirectories, then delete the directory.
+     * Logs a warning if the deletion fails.
      *
      * @param string $path
      * @return bool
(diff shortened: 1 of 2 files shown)


app/Util/storage/safeDeleteDirectory.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace App\Util\Storage;
```
dansup, line 3: The parent directory should be `Storage` (like app/Util/Storage), not `storage` as that isn't PSR-4 compatible. I don't see any usage of this in this PR, can we safely close it now you made #6377

dansup (changes requested):

app/Util/storage/safeDeleteDirectory.php
shleeable: this is pointless now. fastly might have fixed the issue.

Issue #6373, opened 2025-11-23
[Web Bug]: Forgot Email - I can't enter the complete username


joergi:
### Description

I registered a user with the name `PixelfedBugReporter` on `pixelfed.social`
I got the email as a confirmation:   

<img width="515" height="394" alt="Image" src="https://github.com/user-attachments/assets/00ace25a-839e-4a0b-b09a-01b737589a1f" />     

I clicked it, and now I have to wait for the curated onboarding. all good.
  
  
<img width="574" height="699" alt="Image" src="https://github.com/user-attachments/assets/47349105-349f-4f43-b550-aab657df509a" />  
but when I tried to login it said my email is not working, so I clicked the "Forgot Email" -> https://pixelfed.social/auth/forgot/email.  
But I can't enter the full name in this input field, as it's limited to `maxlength="15"`

```html
<input id="username" type="text" class="form-control form-control-lg bg-glass text-white" name="username" maxlength="15" placeholder="Your username" required="">
```

while the registration has 30 letters

<img width="550" height="119" alt="Image" src="https://github.com/user-attachments/assets/a30816ec-da7d-4738-bbe7-ab75d8481da6" />  

```html
<input type="text" class="form-control form-control-lg bg-dark border-dark text-white" placeholder="username" aria-label="Your username" aria-describedby="username-addon" maxlength="30" required="" name="username" value="">
```

### Steps to reproduce

try to enter username PixelfedBugReporter in the page https://pixelfed.social/auth/forgot/email 

### Pixelfed version

v0.12.6

### Acknowledgements

Yes

joergi:
PR for fixing this: https://github.com/pixelfed/pixelfed/pull/6374

shleeable:
PR merged has resolved this.

PR #6374, opened 2025-11-23
uses max length of 30 instead of 15

joergi:
I have also fixed this for the `Admininvite.vue`.

if that's somehow not correct, please revert it, but I guess the limitation makes no sense here too!

Closes issue https://github.com/pixelfed/pixelfed/issues/6373

### DIFF
--- a/resources/views/auth/email/forgot.blade.php
+++ b/resources/views/auth/email/forgot.blade.php
@@ -55,7 +55,7 @@
                                         type="text"
                                         class="form-control form-control-lg bg-glass text-white"
                                         name="username"
-                                        maxlength="15"
+                                        maxlength="30"
                                         placeholder="{{ __('Your username') }}" required>
                                      @if ($errors->has('username') )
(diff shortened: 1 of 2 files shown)


joergi:
Oh I already saw you did from dev to staging.. I changed it and changed it back. 
I hope I haven't caused problems now

resources/views/auth/email/forgot.blade.php
```diff
@@ -55,7 +55,7 @@
                                         type="text"
                                         class="form-control form-control-lg bg-glass text-white"
                                         name="username"
-                                        maxlength="15"
+                                        maxlength="30"
```
rm-yakovenko, line 58: There's a config value that specifies max name length: https://github.com/pixelfed/pixelfed/blob/c68fb2e771d0e1bc5bb1e9eaaaa1e17fb5d3b71f/config/pixelfed.php#L116-L120.

Can we re-use it here to avoid duplication?

dansup (approved):

PR #6378, opened 2025-12-07
Update PushNotificationService.php to include reblogs

grossermensch:
The rest of the code seems to be there, including the formatting, just the type seems to miss.


grossermensch:
@shleeable any comments on that? I tried to follow the notification flow, but the only thing I found missing is the type declaration for reposts.

dansup (approved):

Issue #4044, opened 2022-12-30
Portfolio SQL query missing 'id' in 'group by' clause (Postgres-based installation)


stanclai:
When trying to watch user's portfolio I get an error `column "media.id" must appear in the GROUP BY clause or be used in an aggregate function` in statement `select "status_id" from "media" where "profile_id" = 2 and "status_id" is not null group by "status_id" order by "id" desc limit 50`.

I'm using [latest `dev`](https://github.com/pixelfed/pixelfed/commit/6165f509cf2edd1e326419807c92ef1830dd1cae) with Postgresql database.

Solution: 

```diff
diff --git a/app/Http/Controllers/PortfolioController.php b/app/Http/Controllers/PortfolioController.php
index 5890f2d0..2c00c082 100644
--- a/app/Http/Controllers/PortfolioController.php
+++ b/app/Http/Controllers/PortfolioController.php
@@ -184,7 +184,7 @@ class PortfolioController extends Controller
             return DB::table('media')
             ->whereProfileId($id)
             ->whereNotNull('status_id')
-            ->groupBy('status_id')
+            ->groupBy('status_id', 'id')
             ->orderByDesc('id')
             ->take(50)
             ->pluck('status_id');
```

After applying this patch the problem was resolved. IDK, though, if it's applicable for all cases (I mean, MySQL, for example).

Please see corresponding log lines below:

````
[2022-12-30 07:38:08] production.ERROR: SQLSTATE[42803]: Grouping error: 7 ERROR:  column "media.id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: ...tus_id" is not null group by "status_id" order by "id" desc ...
                                                             ^ (SQL: select "status_id" from "media" where "profile_id" = 2 and "status_id" is not null group by "status_id" order by "id" desc limit 50) {"userId":
2,"exception":"[object] (Illuminate\\Database\\QueryException(code: 42803): SQLSTATE[42803]: Grouping error: 7 ERROR:  column \"media.id\" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: ...tus_id\" is not null group by \"status_id\" order by \"id\" desc ...
                                                             ^ (SQL: select \"status_id\" from \"media\" where \"profile_id\" = 2 and \"status_id\" is not null group by \"status_id\" order by \"id\" desc limit 50
) at /var/lib/pixelfed/vendor/laravel/framework/src/Illuminate/Database/Connection.php:760)
[stacktrace]
[...]

PR #6431, opened 2026-01-22
Fix for portfolio with recent images not showing up on PSQL

grossermensch:
On instances with PostgreSQL database the portfolio stayed empty when selecting „most recent images“ in the portfolio setting. The following error showed up in the web log: 
`'SQLSTATE[42803]: Grouping error: 7 ERROR: column "media.id" must appear in the GROUP BY clause or be used in an aggregate function\nLINE 1: ...tus_id" is not null group by "status_id" order by "id" desc ...\n ^ (Connection: pgsql, SQL: select "status_id" from "media" where "profile_id" = 810030687742828545 and "status_id" is not null group by "status_id" order by "id" desc limit 50)' `

PR introduces a quick fix for the PSQL group by requirements 

Tested on my instance and working now: https://photos.grossermensch.eu/i/portfolio/severin

### DIFF
--- a/app/Http/Controllers/PortfolioController.php
+++ b/app/Http/Controllers/PortfolioController.php
@@ -201,6 +201,7 @@ class PortfolioController extends Controller
             return DB::table('media')
                 ->whereProfileId($id)
                 ->whereNotNull('status_id')
+                ->selectRaw('status_id, MAX(id) as id')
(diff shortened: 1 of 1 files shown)


grossermensch:
Should fix #4044

dansup (approved):

PR #6437, opened 2026-01-26
Update FetchCacheService.php - Cache put/get keys.

shleeable:
```
When the cache contains the key (Cache::has($key) is true), the function returns false instead of retrieving the cached value. Therefore, even if a response were cached, callers never receive it.
The function never stores successful responses in the cache. Only failures are cached (as the integer 1).
```

### DIFF
--- a/app/Services/FetchCacheService.php
+++ b/app/Services/FetchCacheService.php
@@ -18,7 +18,7 @@ class FetchCacheService
         $ar = $allowRedirects ? 'ar1:' : 'ar0';
         $key = self::CACHE_KEY.sha1($url).':'.$vc.$ar.$ttl;
         if (Cache::has($key)) {
-            return false;
+            return Cache::get($key);
         }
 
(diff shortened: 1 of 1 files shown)


dansup (approved):

app/Services/FetchCacheService.php
```diff
@@ -18,7 +18,7 @@ public static function getJson($url, $verifyCheck = true, $ttl = 3600, $allowRed
         $ar = $allowRedirects ? 'ar1:' : 'ar0';
         $key = self::CACHE_KEY.sha1($url).':'.$vc.$ar.$ttl;
         if (Cache::has($key)) {
-            return false;
+            return Cache::get($key);
```
jippi, line 21: why not `::get()` and check the response instead of doing two round-trips (hash and then get)

shleeable: @jippi could you PR this?

PR #6442, opened 2026-02-04
Pull user settings for reblogs in API response 

### LINKED ISSUE #5561: API Timelines don't contain reblogs/boosts
Fetching a user's statuses through the API doesn't show boosted statuses, even though exporting a user's account will contain a list of all the posts the user boosted.

I would expect calling the API would result in the statuses containing the boosts in addition to the user's own statuses.

grossermensch:
Fixes #5561

### DIFF
--- a/app/Http/Controllers/Api/ApiV1Controller.php
+++ b/app/Http/Controllers/Api/ApiV1Controller.php
@@ -2574,7 +2574,13 @@ class ApiV1Controller extends Controller
             $limit = 40;
         }
         $pid = $request->user()->profile_id;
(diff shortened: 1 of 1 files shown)


dansup (approved):

dansup:
@grossermensch Nice work, I just improved this a bit in f76567f67 and will merge to dev shortly.

grossermensch:
@dansup thanks, I try to learn as quickly as I can, but I‘m still very much a beginner 😅 My main goal would be to implement a map view for pixelfed, so images from the server can be found on a map, or searched for based on their location. But I still have a lot to learn until I can tackle that…

PR #6458, opened 2026-02-09
Accept ActivityPub Article type objects

pfefferle:
## Summary

- Add support for incoming `Article` objects ([FEP-b2b8](https://codeberg.org/fediverse/fep/src/branch/main/fep/b2b8/fep-b2b8.md)) so that blog posts from platforms like WordPress, WriteFreely, and NodeBB appear on Pixelfed when they include image attachments
- Route `Article` through the same create, update, and search pipelines as `Note`
- Apply the same domain validation to `Article` URLs as to `Note` URLs

## Changes

- **`app/Util/ActivityPub/Inbox.php`** — accept `Article` in `handleCreateActivity()` (reply and attachment branches) and `handleUpdateActivity()`
- **`app/Util/ActivityPub/Helpers.php`** — include `Article` in domain validation within `storeStatus()`
- **`app/Http/Controllers/SearchController.php`** — accept `Article` in `getPosts()` and `getByUrl()`
- **`app/Services/SearchApiV2Service.php`** — accept `Article` in API v2 search type checks and route it alongside `Note`

## What this does NOT change

- Articles without attachments are still dropped (same as Notes) — Pixelfed is an image platform
- Outgoing posts still emit `Note` — no change to how Pixelfed publishes
- Attachment limits (`MAX_ALBUM_LENGTH`) still apply

## Test plan

- [ ] Search for an Article URL via Pixelfed search — it should resolve and display
- [ ] Incoming Create activity wrapping an Article with image attachments should create a status
- [ ] Update activity for an existing Article should update the status
- [ ] Articles without attachments should still be dropped

### DIFF
--- a/app/Http/Controllers/SearchController.php
+++ b/app/Http/Controllers/SearchController.php
@@ -88,7 +88,7 @@ class SearchController extends Controller
         ) {
             $remote = Helpers::fetchFromUrl($tag);
             if (isset($remote['type']) &&
-                in_array($remote['type'], ['Note', 'Question'])
+                in_array($remote['type'], ['Note', 'Question', 'Article'])
             ) {
                 $item = Helpers::statusFetch($tag);
                 $this->tokens['posts'] = [[
@@ -343,7 +343,7 @@ class SearchController extends Controller
 
         $remote = Helpers::fetchFromUrl($tag);
 
-        if (isset($remote['type']) && $remote['type'] == 'Note') {
+        if (isset($remote['type']) && in_array($remote['type'], ['Note', 'Article'])) {
             $item = Helpers::statusFetch($tag);
             if (! $item) {
                 return;
--- a/app/Services/SearchApiV2Service.php
+++ b/app/Services/SearchApiV2Service.php
@@ -306,7 +306,7 @@ class SearchApiV2Service
                 if ($res) {
                     $json = json_decode($res, true);
 
-                    if (! $json || ! isset($json['@context']) || ! isset($json['type']) || ! in_array($json['type'], ['Note', 'Person'])) {
+                    if (! $json || ! isset($json['@context']) || ! isset($json['type']) || ! in_array($json['type'], ['Note', 'Person', 'Article'])) {
                         return [
                             'accounts' => [],
                             'hashtags' => [],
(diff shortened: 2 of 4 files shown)


dansup:
Blog posts don't belong on photo sharing platforms, IMO

Just because you have a header image, doesn't mean that is main focal point of the context, or one that would be relevant to the Pixelfed community.

Not to mention the spam implications.

pfefferle:
That is security through obscurity I think (no offense). Switching to `Note` on the WordPress site would allow every user to publish images even if they are "only" header images.

The `Note` object type is no indicator for "it is truly an image post" nor is `Article` an indicator for "the post has only a header image".

manton:
To add a concrete example that might help... It's pretty common to have a blog post that has a title, a little text, and a dozen photos. This could be an `Article` but it's really a photo album. Imagine blogging some photos after a trip. It seems like these kind of posts would naturally fit in Pixelfed and other photo platforms.

trwnh:
> indicator for "it is truly an image post"

in a better world this would be the `Image` type >_>

yes, it is weird that pixelfed itself uses Note with attachment. iirc this was uncritically imported from mastodon for more native support in mastodon timelines.

moving forward, i think pixelfed should experiment with better representations of its own data... and if appearance in mastodon timelines is a concern, then this is a problem that should be solved by pressuring mastodon to accept more content than just `Note`

pfefferle:
> moving forward, i think pixelfed should experiment with better representations of its own data... and if appearance in mastodon timelines is a concern, then this is a problem that should be solved by pressuring mastodon to accept more content than just Note

Fighting the same fight, we are fighting with `Article`? ;)

But I think you are right @trwnh, even if it challenging!

pfefferle:
Fun fact: This is btw. exactly how I initially implemented the different WordPress post formats https://github.com/Automattic/wordpress-activitypub/blob/a4887147c6a314931474a1a49d417179f8f98853/includes/model/class-post.php#L153

Not fully semantically correct though 🙂

trwnh:
in a semantic sense, everything is an article... if you mean html:article ;) as in, independent content that can be syndicated and referred to with html:id (like `<article id="foo">`)

that doesn't seem to be how AS2 defined it though x_x

for pixelfed's purpose i think loosely "anything with as:content and as:attachment where each as:attachment has a rdf:type in the set (as:Image as:Video)" is loosely what is being done right now? so it being a Note or Article or anything else at the top level is not really relevant for that. the type check could be removed entirely actually, and pixelfed could simply check for the properties it actually uses instead.

PR #6472, opened 2026-02-18
Update logging.php to rotate laravel.log monthly with 12 months of retention

shleeable:
My laravel.log is 3 years old.


dansup:
The default log driver should be `daily` which automatically rotates the logs.

shleeable:
> The default log driver should be `daily` which automatically rotates the logs.

DOH

PR #6475, opened 2026-02-18
Update PublicApiController.php for findOrFail($username)

shleeable:
`The findOrFail() method expects an integer ID as its parameter, but $username is a string.`

took `$profile = Profile::whereUsername($username)->whereNull('status')->firstOrFail();` from line 93


dansup:
The variable is wrong, the code is working as expected and this PR would break it, we're passing the profile id as a string because of javascript and lack of big int support, but this is indeed the profile_id not the username.

PR #6484, opened 2026-02-18
Safety:  Add defensive check to InboxValidator.php 

shleeable:
This is likely not REQUIRED, but just adds an additional validation.


dansup:
We already do that here

https://github.com/pixelfed/pixelfed/blob/1d3e101bcd1bc509d9d7482d238fa0a587f3db5a/app/Jobs/InboxPipeline/InboxValidator.php#L141

Issue #6485, opened 2026-02-18
[Web Bug]: Resizing of the comment field is not working


joergi:
### Description

If I want to answer a longer text in a comment, it's really hard to read the full text.  
At the moment it has only space for one line.

I tested it in Chromium and Firefox, newest versions, on Linux.

as you can see, the size is not even enough to show 2 lines, if I'm on the comment page and want to comment:
only 1 line is shown completely.
<img width="749" height="104" alt="only one line shown, and no" src="https://github.com/user-attachments/assets/ec70233e-6729-4d9f-b0d2-26677f418aa2" />  

If I'm in the Original Post and want to comment, the `<textarea>` isn't rounded anymore, which is inconsistent, but I can read at least 2 lines - but no handle to resize it either.
<img width="684" height="180" alt="Image" src="https://github.com/user-attachments/assets/e79b9e61-f4b8-4949-9081-06912f11cf74" />

If I want to create a new post, I am able to resize the `<textarea>`
<img width="223" height="169" alt="Image" src="https://github.com/user-attachments/assets/c3005fa1-614e-4a08-9704-226cffbdc3cd" />

the problem is, that the code disallows the resizing ands shows only 3 rows.
```html
<textarea placeholder="Write a comment...." rows="1" .... style="resize: none;...." ></textarea>
```

How it should be:
the size for answering is 2-3 lines.
<img width="733" height="173" alt="Image" src="https://github.com/user-attachments/assets/110115c5-5670-4636-bd0f-da3ce46d3118" />

### Steps to reproduce

1. click on a comment and write in a comment field 2 lines. you can only read one.
2. try to resize the `<textarea>`  . it's not working, no handle to change it.

---

1. click on a post and write in a comment field 2 lines. you can read two lines.
2. try to resize the `<textarea>`  . it's not working, no handle to change it.

fixing: https://github.com/pixelfed/pixelfed/pull/6487

### Pixelfed version

v0.12.7

### Acknowledgements

Yes

joergi:
Okay I just learned, by clicking on accident on the T button, that this makes it bigger.
For me this was always the symbol for changing text font @dansup - so the problem isn't there for writing a comment - but it's there for writing a comment of a comment. 

<img width="198" height="107" alt="Image" src="https://github.com/user-attachments/assets/51845599-b90b-46ff-8c3c-536f5f6359d1" />

I will need to change the PR - I will close the old one, and open a new one!

PR #6487, opened 2026-02-18
Resizing of the comment field is not working - PR for Issue 6485 

joergi:
Commit for bug https://github.com/pixelfed/pixelfed/issues/6485

- it uses 2 rows for textarea
- it removes the resize:none
- makes the textarea also rounded

Fixes pixelfed/pixelfed#6485


joergi:
https://github.com/pixelfed/pixelfed/issues/6485#issuecomment-4055486799

it's not needed as implemented as I just found out by accident.

This will be closed for now, and I will reopen in the future a new one, with just the needed changes

PR #6500, opened 2026-02-25
Fix AccountTransformer.php make sure the acct handle contains the domain suffix for remote users.

shleeable:
`the acct handle lacks the domain suffix (e.g., it shows "alice" instead of "alice@example.com") for remote users.`


dansup (changes requested):
This would break the existing webUI

PR #6519, opened 2026-03-17
Add exif PHP extension to Dockerfile

cgx:
The exif extension is [part of the prerequisites](https://docs.pixelfed.org/running-pixelfed/prerequisites.html) but is not installed in Docker.


shleeable:
This is a good catch. Sorry about this.

dansup (approved):

PR #6555, opened 2026-03-29
Add a config:schema command

### LINKED ISSUE #6168: [Enhancement]: publish schema on configuration options
### Description

It would be nice to offer structured data on available [configuration options](https://docs.pixelfed.org/running-pixelfed/configuration.html), such as using [JSON Schema](https://json-schema.org/).

### Use-case

- generating user interfaces for editing configuration options
- validating configuration validity

### Acknowledgements

Yes

KiaraGrouwstra:
`php artisan config:schema` writes a JSON Schema (draft 2020-12) of every configurable option in the application.

It reads every file in `config/` except `config/schema-meta.php`, which is the command's own overlay. The current tree gives 706 properties. An environment variable sets 685 of them, for 574 distinct variables. 90 are editable at runtime through the admin panel, and 50 hold a secret.

Each property carries the keywords a configuration UI needs:

| keyword | meaning |
|---|---|
| `x-env` | the environment variable that sets the option |
| `x-config-cache` | the option is editable at runtime, not only in the `.env` |
| `x-sensitive` | the value is a secret and must not be shown |
| `x-group` | the section of a form that the option belongs in |

The command evaluates each config file with `env()` replaced by a sentinel, so a default that a function call or a nested `env()` computes still resolves. A tokenizer pass then checks that every `env()` call in every file reached a property, and the command fails if one did not. A new option therefore cannot enter `config/` without the schema seeing it. `config/schema-meta.php` supplies the title, description, group and enum that a config file cannot express, and gives an `env` entry for the few keys the evaluator cannot reach.

A workflow regenerates the schema on each pull request to `dev` and fails if the committed `pixelfed-config.schema.json` is out of date.

Fediversity uses the output to build an operator-facing form for a hosted instance. The schema already found eight option names in that project's table that pixelfed does not read, or reads only as a deprecated fallback.

closes #6168

Assisted-by: Claude:claude-fable-5-1

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -33,6 +33,16 @@ jobs:
           cp .env.testing .env
           php artisan key:generate
 
+      - name: Check config schema is up to date
+        run: |
+          php artisan config:schema --pretty --output=/tmp/schema-check.json
+          if ! diff -q pixelfed-config.schema.json /tmp/schema-check.json > /dev/null 2>&1; then
+            echo "::error::pixelfed-config.schema.json is out of date."
+            echo "Run 'php artisan config:schema --pretty --output=pixelfed-config.schema.json' and commit the result."
+            diff pixelfed-config.schema.json /tmp/schema-check.json
+            exit 1
+          fi
+
       - name: Run tests
         run: php artisan test
         continue-on-error: false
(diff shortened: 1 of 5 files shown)


shleeable:
Dan and I have been discussing the .env and configcache stuff earlier today.

I plan on doing some refactoring around that, and then I will try to get this merged in. It is one of those features that might only be used for a small handful of people, but having a trusted schema for the configuration sounds like a step in the right direction.

I will try to get my refactoring completed first and then look into this.

PR #6556, opened 2026-03-30
Migrate to Intervention Image v4

olivervogel:
This PR makes the necessary changes to migrate to the just released [Intervention Image Version 4](https://github.com/Intervention/image/releases/tag/4.0.0). 

Version 4 is more of a cosmetic update that primarily improved the public API.

I might have missed something, but I think this is a good start and takes care of most of the work for the update. I recommend a thorough review.

If you have any questions, feel free to reach out to me.

PS: For your reference, here is the [upgrade guide](https://image.intervention.io/v4/getting-started/upgrade)


shleeable:
Thank you

shleeable:
This work will be completed in another PR to update the composer lock... Thank you for your assistance

PR #6557, opened 2026-04-01
Fix profile cache not clearing when updating only pronouns

Gredelston:
### What does this PR do?
Fixes an issue where a user's profile cache is not cleared if they only update their pronouns in settings. 

### Why is this needed?
When a user updates their pronouns without modifying any other profile fields (like name, bio, or website), the `$changes` flag in `HomeSettings.php` remains `false`. As a result, the backend correctly saves the new pronouns to the database via `PronounService::put()`, but it skips clearing the profile cache. This causes the old pronouns to persist on the frontend until the cache expires or another field is updated.

This PR sets `$changes = true` when pronouns are modified, ensuring the cache is properly invalidated.

### How to test
1. Go to Account Settings > Profile.
2. Update your pronouns, leaving all other fields unchanged.
3. Save changes.
4. Verify that the new pronouns are immediately visible on your profile page.

### DIFF
--- a/app/Http/Controllers/Settings/HomeSettings.php
+++ b/app/Http/Controllers/Settings/HomeSettings.php
@@ -86,6 +86,7 @@ trait HomeSettings
             }
 
             if ($existingPronouns != $pronouns) {
+                $changes = true;
                 if ($pronouns && in_array('Select Pronoun(s)', $pronouns)) {
                     PronounService::clear($profile->id);
                 } else {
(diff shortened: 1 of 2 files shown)


Gredelston:
I've added a second commit that fixes the rendering of pronouns on the profile page. It was missing from the recently refactored ProfileSidebar.vue component.

shleeable:
Thank you for this. please note, Daniel is rewriting the entire of the frontend.. so that code might not actually be used in the next release.

PR #6589, opened 2026-04-24
Accept compacted `Note` attachments

### LINKED ISSUE #6588: [Federation]: `Create(Note)` with a single attachment is dropped because `Helpers::verifyAttachments` still assumes `attachment` is an array (follow-up to #6427)
### Description

PR #6427 made `handleCreateActivity` accept the JSON-LD 1.1 compact form of `to`/`cc` (single value instead of a one-element array).  The same compaction rule applies to every set-valued property in Activity Streams 2.0, and one of those properties—`attachment`—is still validated with Laravel's array-shape rules, which silently reject any payload where `attachment` is a single object.

The concrete symptom: a signed `Create(Note)` delivery with one image attachment is accepted over HTTP (2xx), but the note never shows up in the Home Feed, Global Feed, or on the author's remote profile.  `verifyNoteAttachment()` returns `false`, so `handleNoteCreate` is skipped and the activity is dropped silently.  This reproduces with every sender I've tested that uses a standards-compliant JSON-LD compactor (Fedify 2.1, and by extension anything built on [`jsonld`][npm-jsonld]); it does not reproduce with Pixelfed→Pixelfed traffic because Pixelfed's outgoing `Note` transformer always emits an array.

I'm happy to open a PR mirroring the style of #6427 (a local `is_array`/wrap-in-array normalisation inside `verifyAttachments`) if that's the preferred approach.

[npm-jsonld]: https://www.npmjs.com/package/jsonld

### Steps to reproduce

A minimal payload that Pixelfed's shared inbox receives from a Fedify-backed server (signatures omitted):

```json
{
  "@context": ["https://www.w3.org/ns/activitystreams"],
  "id": "https://remote.example/users/alice#posts/abc/create",
  "type": "Create",
  "actor": "https://remote.example/users/alice",
  "to": "https://www.w3.org/ns/activitystreams#Public",
  "cc": "https://remote.example/users/alice/followers",
  "object": {
    "id": "https://remote.example/users/alice/posts/abc",
    "type": "Note",
    "attributedTo": "https://remote.example/users/alice",
    "content": "<p>Hello!</p>",
    "attachment": {
      "type": "Document",
      "mediaType": "image/png",
      "url": "https://remote.example/uploads/abc.png"
    }
  }
}
```

 -  `to`/`cc` pass because of #6427.
 -  `attachment` is a single `Document` object, which `Helpers::verifyAttachments` rejects.
 -  The same delivery with `"attachment": [{...}]` (one-element array) renders correctly.

### Root cause

`Inbox::handleCreateActivity()` guards the note branch with `Inbox::verifyNoteAttachment()`:

https://github.com/pixelfed/pixelfed/blob/0e07fd744ccfdabda7a88b37cf82c348694a67ba/app/Util/ActivityPub/Inbox.php#L279-L284

`Inbox::verifyNoteAttachment()` → `Helpers::verifyAttachments` uses Laravel's `*.field` rule:

https://github.com/pixelfed/pixelfed/blob/0e07fd744ccfdabda7a88b37cf82c348694a67ba/app/Util/ActivityPub/Helpers.php#L90-L98

`*.field` only iterates when the input is a plain array.  A single `Document` object arrives as a map, `*` can't iterate it, and validation fails.  This is exactly the same pre-#6427 failure mode the `to`/`cc` patch addressed—it just hasn't been extended to `attachment`.

### Why this is a spec issue, same as #6427

The [Activity Streams 2.0 Core spec, §4.6 *Properties with Multiple Values*][as2-core-set] says:

> For properties that can take multiple values, the JSON serialization MAY either be a single value or an array.

[JSON-LD 1.1 compact form][jsonld-1.1-compact] formalizes the identical behavior: unless a term is tagged `@container: "@set"` in the context, compact output collapses single-value arrays into the value alone.  `attachment` is not a `@set` term in the AS2 context, so every compliant JSON-LD compactor will drop the array wrapper for one-element inputs.  #6427 already recognized this for `to`/`cc`.

Mastodon, Misskey, and GoToSocial all handle this without special-casing each property because they run incoming payloads through a JSON-LD expand step and validate on the expanded form.  Pixelfed's inbox validates structurally on the compact form, which is why the fix has to be property-by-property unless/until an expand pass is introduced.

[as2-core-set]: https://www.w3.org/TR/activitystreams-core/#h-array
[jsonld-1.1-compact]: https://www.w3.org/TR/json-ld/#compacted-document-form

### Proposed fix (matching the style of #6427)

The smallest change, and the one closest to the existing `to`/`cc` mitigation, is a local normalisation inside `Helpers::verifyAttachments`:

```php
public static function verifyAttachments(array $data): bool
{
    if (! isset($data['object']) || empty($data['object'])) {
        $data = ['object' => $data];
    }

    $activity = $data['object'];
    // ...

    if (! isset($activity['attachment']) || empty($activity['attachment'])) {
        return false;
    }

    $attachment = $activity['attachment'];
    // JSON-LD allows one-element arrays to collapse to the value alone.
    // Normalise so the '*.*' validator rules below can iterate.
    if (is_array($attachment) && ! array_is_list($attachment)) {
        $attachment = [$attachment];
    }

    return Validator::make($attachment, [
        '*.type' => ['required', 'string', Rule::in($mediaTypes)],
        // ...
    ])->passes();
}
```

(`array_is_list()` is available in PHP 8.1+, which matches Pixelfed's current minimum.)

Once this is in place, the `Note` transformer in `handleNoteCreate()` also needs to treat `attachment` as a possibly-single-or-array value; happy to include that in the same PR.

### Other places the same normalization is worth applying

While checking with `grep`, the same pattern assumption shows up in a few more validators / helpers.  None of them are as common as attachment, but they'll bite the next sender:

 -  `Accept::validate()` (https://github.com/pixelfed/pixelfed/blob/dev/app/Util/ActivityPub/Validator/Accept.php) requires `object.actor` and `object.object` as URL strings.  A sender that compacts the inner `Follow` with an inlined `actor` (which Fedify happens to do when `follow.getActor()` was called during inbox processing) is rejected.
 -  `Helpers::normalizeAudience()` (https://github.com/pixelfed/pixelfed/blob/0e07fd744ccfdabda7a88b37cf82c348694a67ba/app/Util/ActivityPub/Helpers.php) only handles `to`/`cc` as arrays; when they arrive as single strings outside of `Inbox::handleCreateActivity()`'s mitigation path, the method returns `null`.
 -  `tag` parsing in a couple of transformers (`Status`, mentions) assumes array form.

Each of these could either get a local normalizer like the one above, or—cleaner—the inbox could expand the payload through a JSON-LD processor once at the top and validate against the expanded form.

### Pixelfed version

0.12.5

### Acknowledgements

Yes

dahlia:
Fixes incoming ActivityPub `Create(Note)` deliveries where `attachment` is encoded as a single compacted object instead of a one-item array.

Pixelfed now normalizes Note attachments before validation and import, so single JSON-LD compacted attachments are handled the same way as existing array-form attachments.

Fixes #6588.

### DIFF
--- a/app/Util/ActivityPub/Helpers.php
+++ b/app/Util/ActivityPub/Helpers.php
@@ -77,17 +77,18 @@ class Helpers
             $data = ['object' => $data];
         }
 
-        $activity = $data['object'];
         $mimeTypes = explode(',', config_cache('pixelfed.media_types'));
         $mediaTypes = in_array('video/mp4', $mimeTypes) ?
             ['Document', 'Image', 'Video'] :
             ['Document', 'Image'];
 
-        if (! isset($activity['attachment']) || empty($activity['attachment'])) {
+        $attachments = self::getAttachments($data);
+
+        if (empty($attachments)) {
             return false;
         }
 
-        return Validator::make($activity['attachment'], [
+        return Validator::make($attachments, [
             '*.type' => ['required', 'string', Rule::in($mediaTypes)],
             '*.url' => 'required|url',
             '*.mediaType' => ['required', 'string', Rule::in($mimeTypes)],
@@ -901,9 +902,21 @@ class Helpers
      */
     public static function getAttachments(array $data): array
     {
-        return isset($data['object']) ?
-            $data['object']['attachment'] :
-            $data['attachment'];
+        $object = isset($data['object']) ?
+            $data['object'] :
+            $data;
+
+        if (! is_array($object) ||
+            ! isset($object['attachment']) ||
+            empty($object['attachment']) ||
+            ! is_array($object['attachment'])
+        ) {
+            return [];
+        }
+
+        return array_is_list($object['attachment']) ?
+            $object['attachment'] :
+            [$object['attachment']];
     }
 
     /**
--- a/tests/Unit/ActivityPub/NoteAttachmentTest.php
+++ b/tests/Unit/ActivityPub/NoteAttachmentTest.php
@@ -56,5 +56,56 @@ class NoteAttachmentTest extends TestCase
         $valid = Helpers::verifyAttachments($this->invalidMime);
         $this->assertFalse($valid);
     }
-}
 
+    #[Test]
+    public function compactedCreateNoteAttachment()
+    {
+        $activity = $this->mastodon;
+        $activity['object']['attachment'] = $activity['object']['attachment'][0];
+
+        $valid = Helpers::verifyAttachments($activity);
+        $this->assertTrue($valid);
+    }
+
+    #[Test]
+    public function compactedBareNoteAttachment()
+    {
+        $activity = $this->pixelfed;
+        $activity['attachment'] = $activity['attachment'][0];
+
+        $valid = Helpers::verifyAttachments($activity);
+        $this->assertTrue($valid);
+    }
+
+    #[Test]
+    public function compactedInvalidAttachmentType()
+    {
+        $activity = $this->invalidType;
+        $activity['object']['attachment'] = $activity['object']['attachment'][0];
+
+        $valid = Helpers::verifyAttachments($activity);
+        $this->assertFalse($valid);
+    }
+
+    #[Test]
+    public function compactedInvalidMimeType()
+    {
+        $activity = $this->invalidMime;
+        $activity['object']['attachment'] = $activity['object']['attachment'][0];
+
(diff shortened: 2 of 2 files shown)


shleeable:
Closed in PR #6950

thank you for your support and resolving this issue.

PR #6602, opened 2026-05-09
Media: support .mov uploads via on-ingest transcode to mp4

nosguy:
## Problem

The admin Media settings UI exposes a "MOV" checkbox that adds the string `video/mov` to `pixelfed.media_types`. `video/mov` is not a registered IANA mime type — PHP's `finfo` (which is what `UploadedFile::getMimeType()` calls) returns `video/quicktime` for actual `.mov` files. So when an admin enables MOV support, every real `.mov` upload arrives with mime `video/quicktime`, fails the `mimetypes:` validator at `ComposeController.php:106` / `ApiV2Controller.php:259`, and is rejected with 400/403 “Invalid media format.”

Even if the validation were patched to accept `video/quicktime`, the post-upload switch in both controllers only dispatches `VideoThumbnail` for `video/mp4` — quicktime media would land in the DB without a thumbnail and fail to publish.

## Fix

Two halves:

1. **Use the correct mime strings.** Replace `video/mov` with `video/quicktime` in the admin allowlist (`AdminSettingsController` validator + `AdminSettings.vue` writer), and add `video/x-m4v` for `.m4v`. The Vue `setMediaTypes` parser is rewritten as an explicit mime→key map so existing cached `video/mov` values still toggle the MOV checkbox correctly (backwards compat). `StoryFetch` mime→extension table updated similarly.

2. **New `App\Jobs\VideoPipeline\VideoTranscode` job.** Opens the source via the existing `pbmedia/laravel-ffmpeg` integration, exports H.264/AAC mp4 with `-movflags +faststart`, swaps `media_path`/`mime`/`size`, deletes the original, then chains into `VideoThumbnail`. `ComposeController` and `Api\ApiV2Controller` route `video/quicktime` and `video/x-m4v` uploads to it.

No migrations, no config changes required. Existing instances with `video/mov` in their cached `media_types` keep working at the admin UI parser level until rewritten.

## Verified

Deployed and tested on a self-hosted instance: an iPhone HEVC `.mov` upload now succeeds end-to-end — transcode runs on the `mmo` queue, thumbnail and blurhash generate, and the post publishes and plays in the timeline.

## Files changed

- `app/Jobs/VideoPipeline/VideoTranscode.php` (new)
- `app/Http/Controllers/ComposeController.php` (case for quicktime/x-m4v)
- `app/Http/Controllers/Api/ApiV2Controller.php` (case for quicktime/x-m4v)
- `app/Http/Controllers/Admin/AdminSettingsController.php` (allowlist)
- `resources/assets/components/admin/AdminSettings.vue` (writer + parser)
- `app/Jobs/StoryPipeline/StoryFetch.php` (mime→ext table)

### DIFF
--- a/app/Http/Controllers/Admin/AdminSettingsController.php
+++ b/app/Http/Controllers/Admin/AdminSettingsController.php
@@ -603,7 +603,7 @@ trait AdminSettingsController
         $mediaTypes = $request->input('media_types');
         $mediaArray = explode(',', $mediaTypes);
         foreach ($mediaArray as $mediaType) {
-            if (! in_array($mediaType, ['image/jpeg', 'image/png', 'image/gif', 'image/webp', 'image/avif', 'image/heic', 'video/mp4', 'video/mov'])) {
+            if (! in_array($mediaType, ['image/jpeg', 'image/png', 'image/gif', 'image/webp', 'image/avif', 'image/heic', 'video/mp4', 'video/quicktime', 'video/x-m4v'])) {
                 return redirect()->back()->withErrors(['media_types' => 'Invalid media type']);
             }
         }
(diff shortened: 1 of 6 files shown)


shleeable:
Hello, we've recently upgraded to the latest VIPs package so this needs to be refactored regardless but...

we're also looking into client side video transcoding as well.. or moving to ffmpeg for transcoding video.

We will look into this in the future.

Issue #6604, opened 2026-05-11
[Web Bug]: Missing Spanish translations


raymundovr:
### Description

The Spanish translation is missing different content accross the application.

### Steps to reproduce

For example, go to Settings > Filters. All the content there is missing a translation into this language

### Pixelfed version

0.12.7

### Acknowledgements

Yes

raymundovr:
Hi,
I've just closed my previous PR after noticing that it might not be the right place to make the modifications.

Could you confirm that the place where the translations need to be set is `resources/lang/es/*.php`?

How can I regenerate the corresponding `.json` translations? I have tried `npm run development` and `npm run production` without success.

Thanks!

shleeable:
Hello, thank you for assisting with translation. We will aim to try and improve the language support

shleeable:
Actually. We will be rebuilding a new WebUI with an updated translation system.

shleeable:
Hello, the translation system has been refactored.

crowdin is now the source of truth for transcription.

raymundovr:
@shleeable  Understood! Thanks! I'm checking the platform right now.

raymundovr:
I've spotted this issue where I cannot save the translation due to the placeholder, or is just bad UI feedback?
<img width="1044" height="108" alt="Image" src="https://github.com/user-attachments/assets/cb762738-be5e-4449-9915-f1840ed5a782" />

raymundovr:
@shleeable translations added there :)

PR #6605, opened 2026-05-11
Improve Spanish translation

raymundovr:
This PR addresses content throughout the application that is missing an Spanish translation.
It also aligns a couple of sentences with previously translated content.

See #6604


shleeable:
Hello, thank you for assisting with translation.

PR #6617, opened 2026-05-23
fix: return 422 (not 500) for ValidationException on JSON API requests

### LINKED ISSUE #6609: [Server Bug]: API JSON responses return HTTP 500 for ValidationException instead of 422
### Summary
  The global exception handler returns HTTP **500** for every Laravel `ValidationException`
  on JSON API requests, instead of the expected **422 Unprocessable Entity**.

  This makes every Mastodon-compatible client treat ordinary validation failures
  as server errors, which surfaces to end-users as "Oops! The service may be
  temporarily unavailable" rather than a recoverable client error.

  ### Root cause
  `app/Exceptions/Handler.php` lines 68–82 (current `dev`, fdeb29ba8):

      if ($exception instanceof \Illuminate\Validation\ValidationException && $request->wantsJson()) {
          return response()->json(
              [
                  'message' => $exception->getMessage(),
                  'errors'  => $exception->validator->getMessageBag(),
              ],
              method_exists($exception, 'getStatusCode') ? $exception->getStatusCode() : 500
          );
      }

  `ValidationException` does **not** define `getStatusCode()` — it exposes a
  public `$status` property (default `422`). The `method_exists()` check therefore
  always evaluates to `false`, and every validation failure falls through to the
  `500` literal.

  This regressed in 2022 (commit `74905ba1d`) when the response was switched from
  `->setStatusCode($exception->status)` to the current ternary.

  ### Suggested fix
  Replace the ternary with `$exception->status ?? 422`. Happy to open a PR.

  ### Environment
  - Pixelfed `dev` @ fdeb29ba8 (also reproduced on 0.12.7)
  - Affects all Mastodon-API clients hitting any `$this->validate(...)` endpoint

### Reproduction
  Any API endpoint that uses `$this->validate(...)` and fails validation.
  Smallest repro against a running instance:

      curl -i -H "Authorization: Bearer <token>" \
        "https://<instance>/api/v1/timelines/home?max_id="

  Observed: `HTTP/1.1 500 Internal Server Error`
  Expected: `HTTP/1.1 422 Unprocessable Entity`

  The JSON body is otherwise correct (it contains `message` and `errors`), so the
  status code is the only thing wrong.

### Pixelfed version

0.12.7

### Acknowledgements

Yes

### LINKED ISSUE #6610: [Server Bug]: /api/v1/timelines/home rejects empty max_id=/min_id= parameters (Mastodon-compat clients can't paginate)
### Summary
  `GET /api/v1/timelines/home?max_id=` (empty value) fails validation with
  HTTP 422 (or 500 today — see #<issue 1>), while every other Mastodon-API
  timeline endpoint accepts an empty `max_id`/`min_id` without complaint.

  Several Mastodon-API clients — including Pixelfed for iOS — send
  `max_id=` on the first page load. The Pixelfed iOS app surfaces this as a
  "Stories" load failure because the home timeline call happens alongside the
  stories fetch.

### Root cause
  `app/Http/Controllers/Api/ApiV1Controller.php::timelineHome()`
  (lines 2608–2614 on `dev` @ fdeb29ba8):

      $this->validate($request, [
          'page'           => 'sometimes|integer|max:40',
          'min_id'         => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
          'max_id'         => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
          'limit'          => 'sometimes|integer|min:1',
          'include_reblogs'=> 'sometimes',
      ]);

  The `ConvertEmptyStringsToNull` middleware turns `?max_id=` into `null`.
  With `sometimes|integer`, the field is *present* (so `sometimes` doesn't skip
  it) but `null` is not an `integer`, so validation fails.

  Every other timeline/listing endpoint in this same file uses
  `nullable|integer|min:0|max:`PHP_INT_MAX — see e.g. `timelinePublic`,
  `accountStatusesById`, `timelineHashtag`. `timelineHome` is the outlier.

  ### Suggested fix
  Change `sometimes` → `nullable` for `min_id` and `max_id` (and arguably `limit`
  / `page`) so behaviour matches the rest of the controller. Happy to open a PR.

  ### Environment
  - Pixelfed `dev` @ fdeb29ba8 (also 0.12.7)
  - Client: Pixelfed for iOS, also reproducible with curl

### Steps to reproduce

   curl -i -H "Authorization: Bearer <token>" \
        "https://<instance>/api/v1/timelines/home?max_id="

  Returns 422 (or 500 if #<issue 1> is in effect) with:

      {"message":"The max id must be an integer.","errors":{"max_id":["The max id must be an integer."]}}

  Same request against `/api/v1/timelines/public` or `/api/v1/accounts/:id/statuses`
  returns 200 — those endpoints already use `nullable|integer`.

### Pixelfed version

0.12.7

### Acknowledgements

Yes

cryptomentor-de:
Fixes #6609, fixes #6610

## Summary

Two closely related bugs that together made the Pixelfed Android app (and other Mastodon-compatible clients) receive HTTP 500 on every home timeline request.

### Bug 1 — Handler.php returns HTTP 500 for ValidationException (#6609)

`ValidationException` has no `getStatusCode()` method — it only has a `$status` property (default `422`). The existing handler checked `method_exists($exception, 'getStatusCode')`, which returns `false`, so the fallback `500` was used for every JSON API validation failure.

```php
// Before
method_exists($exception, 'getStatusCode') ? $exception->getStatusCode() : 500

// After
method_exists($exception, 'getStatusCode') ? $exception->getStatusCode() : $exception->status
```

### Bug 2 — `timelineHome` rejects empty `max_id=` / `min_id=` (#6610)

The Pixelfed Android app (okhttp/4.12.0) sends `max_id=` (empty string) on every first home timeline load. The `ConvertEmptyStringsToNull` middleware converts this to `null` before validation reaches the controller. The rule `'sometimes|integer'` still validated the key (it is "present" as `null`) and the `integer` rule failed on `null`, throwing a `ValidationException` — which then triggered bug #6609, returning HTTP 500.

Every other timeline method in `ApiV1Controller` already uses `'nullable|integer'` for `max_id`/`min_id`. This PR makes `timelineHome` consistent.

```php
// Before
'min_id' => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
'max_id' => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,

// After
'min_id' => 'nullable|integer|min:0|max:'.PHP_INT_MAX,
'max_id' => 'nullable|integer|min:0|max:'.PHP_INT_MAX,
```

## Affected clients

Any Mastodon-compatible client that sends `max_id=` or `min_id=` as an empty string on the first page load. Confirmed with:
- Pixelfed Android app (okhttp/4.12.0)
- Pixelfed iOS app (stories load alongside the home timeline call)

## Tests

`tests/Feature/HomeTimelineTest.php` adds four regression tests:

1. Non-integer `max_id` returns 422 (not 500) — isolates the Handler.php fix
2. Empty `max_id=` returns a successful response — the exact request the Android app sends
3. Same for empty `min_id=`
4. Non-null non-integer `max_id` is still rejected with 422 — the `nullable` change must not weaken input validation

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -71,7 +71,7 @@ class Handler extends ExceptionHandler
                     'message' => $exception->getMessage(),
                     'errors' => $exception->validator->getMessageBag(),
                 ],
-                method_exists($exception, 'getStatusCode') ? $exception->getStatusCode() : 500
+                method_exists($exception, 'getStatusCode') ? $exception->getStatusCode() : $exception->status
             );
         } elseif ($request->wantsJson()) {
             return response()->json(
--- a/app/Http/Controllers/Api/ApiV1Controller.php
+++ b/app/Http/Controllers/Api/ApiV1Controller.php
@@ -2607,8 +2607,8 @@ class ApiV1Controller extends Controller
 
         $this->validate($request, [
             'page' => 'sometimes|integer|max:40',
-            'min_id' => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
-            'max_id' => 'sometimes|integer|min:0|max:'.PHP_INT_MAX,
+            'min_id' => 'nullable|integer|min:0|max:'.PHP_INT_MAX,
+            'max_id' => 'nullable|integer|min:0|max:'.PHP_INT_MAX,
             'limit' => 'sometimes|integer|min:1',
             'include_reblogs' => 'sometimes',
         ]);
--- a/tests/Feature/HomeTimelineTest.php
+++ b/tests/Feature/HomeTimelineTest.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace Tests\Feature;
+
+use App\User;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Laravel\Passport\Passport;
+use PHPUnit\Framework\Attributes\Test;
+use Tests\TestCase;
+
+/**
+ * Regression tests for two related bugs that together caused HTTP 500 on the
+ * home timeline endpoint for Mastodon-compatible clients (including the
+ * Pixelfed Android app).
+ *
+ * Bug 1 — Issue #6609: Handler.php returns HTTP 500 for ValidationException
+ * Bug 2 — Issue #6610: timelineHome rejects empty max_id= / min_id= parameters
+ */
+class HomeTimelineTest extends TestCase
+{
+    use RefreshDatabase;
+
+    private function authenticatedUser(): User
+    {
+        $user = User::factory()->create();
+        Passport::actingAs($user, ['read']);
+
+        return $user;
+    }
+
+    // -------------------------------------------------------------------------
+    // Bug #6609 — Exception handler returns 500 instead of 422
+    // -------------------------------------------------------------------------
+
+    /**
+     * ValidationException must produce HTTP 422, not 500.
+     *
+     * Root cause: Handler.php called method_exists($exception, 'getStatusCode')
+     * which is false for Illuminate\Validation\ValidationException (the class
+     * exposes $status = 422 but has no getStatusCode() method), so the fallback
+     * of 500 was used for every JSON API validation failure.
+     */
+    #[Test]
+    public function json_api_validation_failure_returns_422_not_500(): void
+    {
+        $this->authenticatedUser();
+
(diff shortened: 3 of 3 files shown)


dansup:
This was already fixed! Thanks though <3

PR #6618, opened 2026-05-23
fix: apply EXIF orientation before resizing images

cryptomentor-de:
## Problem

Portrait photos uploaded from smartphones are displayed as landscape after processing.

**Root cause:** `handleImageTransform()` in `app/Util/Media/Image.php` calls `$img->width()` and `$img->height()` immediately after `read()`, without first applying the EXIF orientation tag. Smartphones store photos in the sensor's native orientation (typically landscape) and set an EXIF `Orientation` tag (e.g. `6` = rotate 90° CW) to indicate how the image should be displayed.

Since the raw pixel dimensions are used to determine aspect ratio and scaling, a portrait photo shot at 3024×4032 (physical) with `Orientation=6` is classified as landscape and resized to 1920×1080. The EXIF tag is not preserved in the output, so the result is permanently wrong.

## Fix

Call `orient()` immediately after `read()`. Intervention Image v3's `orient()` reads the EXIF orientation tag, physically rotates the image to match, and clears the tag — so all subsequent dimension queries and resize operations work on the correctly oriented image.

```php
$img = $this->imageManager->read($fileContents);
$img = $img->orient(); // apply EXIF rotation before dimension checks
```

## Testing

1. Upload a portrait photo taken with a smartphone (EXIF Orientation ≠ 1)
2. Verify the processed image in storage is in portrait orientation
3. Verify the thumbnail is also in portrait orientation

## References

- Intervention Image v3 `orient()`: https://image.intervention.io/v3/modifying/orientation


lcx:
@cryptomentor-de thanks for this, tried your fix on my pixelfed, rebuilt and restarted and to base safe also ran 

```
php artisan config:cache
php artisan cache:clear
php artisan queue:restart
```

but nothing changed. When using the Pixelfed app and uploading from an iPhone I still have the same issue.

shleeable:
I will update to the latest Intervention Image release and try this PR again.

Issue #4486, opened 2023-06-17
Support HDR image formats (like JXR)


fofwisdom:
Windows 11 supports JXR and produces a native JXR. It has quickly become popular. It would be great to have a roadmap for JXR support in Pixelfed.

trwnh:
JPEG XR is not "popular"; it is also no longer supported and has been superseded by JPEG XL.

fofwisdom:
> JPEG XR is not "popular"; it is also no longer supported and has been superseded by JPEG XL.

I'm surprised. 
Thanks for the feedback and advice.
Microsoft & Nvidia has been coding for JXR support for recently years. All Windows 11 users get JXR when they press to capture a screen in HDR situations.
Does this mean that Microsoft will also be switching support to JXL?

fofwisdom:
JXL has also been abandoned, so we'll have to wait for WebP 2. HDR will never see the light of day in the web world while JXR stacks up locally on Windows users for years to come.
Too bad.
Thanks.

trwnh:
I'm not sure what Microsoft will be doing, but I guess for Pixelfed the best we can do is maybe see if it can be converted by gd or imagick. But I suspect if it's already not working, then they likely don't support it. There's also a config option for which file types are allowed to be uploaded, but it's what happens after upload that matters.

I can reopen this issue as a more general request for HDR image support, but I suspect that we will be waiting for a while until the tooling supports it... Of the HDR formats listed in https://en.wikipedia.org/wiki/Category:High_dynamic_range_file_formats I see AVIF and HEIF as the only ones that are likely to be supported. The bottleneck is in the image processor running on the Pixelfed host server -- putting aside GD which only supports some basic formats, ImageMagick supports a lot of formats but not every format: https://imagemagick.org/script/formats.php#supported

In theory JXR is listed there but it depends on `jxrlib`. So it is possible that it could work, but only under the following requirements:

- The admin has configured the server to use `imagick` instead of `gd`
- The admin has installed `jxrlib` as an extra dependency
- The admin has allowed upload of `image/vnd.ms-photo, image/jxr` MIME types

after which, the image would probably be converted to a normal JPEG for web viewing, and in doing so it loses HDR information probably?

fofwisdom:
My thoughts were short.

It seems like the market and the community are going their own way without reaching a consensus. There is also the issue of the direction of image standards by web browser engine maintainers.

[...]

Issue #3102, opened 2021-12-29
JPEG XL support


xplosionmind:
Even though it is not yet widely supported, [JPEG XL](https://jpeg.org/jpegxl/ 'JPEG XL official website') appears to be game-changing and it offers unprecedented quality and compression.

If Pixelfed implemented it using the already available [PHP library](https://github.com/joppuyo/jpeg-xl-encode 'A PHP library for encoding JPEG XL images — GitHub') to handle it, it would be extremely appealing cutting-edge tech that very few mainstream systems support yet.

More info:
- [standard library](https://github.com/libjxl/libjxl)
- [specification](https://jpeg.org/jpegxl/documentation.html)
- [how it crushes other formats](https://cloudinary.com/blog/how_jpeg_xl_compares_to_other_image_codecs)

LeoEurope:
I would like to suggest considering JPEG-XL (.jxl) as the only internal image format, not just as an also-supported-format.

### The benefits for Pixelfed include:

* It has the best image quality at the smallest size of all JPEG successors (JPEG2000, WebP, AVIF, HEIC).
* It incorporates a lossless mode that outperforms PNG compression levels.
* It can transcode existing JPEG files to further reduce their size without reducing image quality.
* It supports progressive image delivery, useful for slower connections. Smaller files transfer faster too.
* It encodes faster than many other modern image formats.
* It is the favourite format of the photography space because of its unparalleled colour depth (up to 32 bits, HDR & Wide Gamut), practically unlimited dimensions (approximately 1 billion × 1 billion pixels) and transparency (alpha channel).
* It is royalty-free with a BSD-3 license and an open standard defined by ISO/IEC 18181, developed by the Joint Photographic Experts Group (JPEG), Google and Cloudinary.
* It could set Pixelfed apart as the social network that truly cares about photo quality.
* JPEG and the JPEG-XL cheerleaders would promote Pixelfed as an example of adoption.

Making JPEG-XL central to all image handling would also close a number of somewhat related requests such as:
https://github.com/pixelfed/pixelfed/issues/4486 , https://github.com/pixelfed/pixelfed/issues/5159 and a number of requests for other image formats that are not (fully) supported. As JPEG-XL's format abilities extend beyond what other image formats support, converting submitted JPEG, WebP, AVIF, HEIC etc. to JPEG-XL can happen without information loss. After that you only have to worry about rendering JPEG-XL, not any of the other formats.

### How I see the implementation:

[...]

PR #6650, opened 2026-07-07
[HDR] Use libvips for converting to specified storage format

dshatz:
Draft!

Following discussions in #4486 and #3102, this is my idea of how it could work.

Utilizing libvips for image manipulation, `ImageOptimize` job will convert all uploaded images to one configured format (`IMAGE_STORAGE_FORMAT`). 

The main benefit is better HDR handling - whatever the users upload, HDR will be preserved and served in either JPEG with ISO gainmap or JXL.

I tested this by uploading the following formats and checking that the optimized image is a valid JPEG with ISO gainmap:
 - JPEG with ISO gainmap
 - JPEG with ISO and XMP gainmap
 - AVIF with ISO gainmap
 - HDR AVIF
 - HDR HEIC
 
 For the last two, libvips needs to be built from master (necessary changes are scheduled for 8.19 in the coming months).
 
 _Theoretically JXL uploads should also work but I couldn't make it work because mimeTypes are hardcoded all over the place._


dshatz:
Asking for a review so I know if this is an acceptable direction.

shleeable:
Hello, we've recently upgraded to the latest VIPs package so this needs to be refactored regardless but...

We are considering transcoding client side to a few known formats. 
also interested in making use of png, webp or even jpeg-xl

dshatz:
> Hello, we've recently upgraded to the latest VIPs package so this needs to be refactored regardless but...
> 
> We are considering transcoding client side to a few known formats. also interested in making use of png, webp or even jpeg-xl

I think I don't understand - are you saying that transcoding will be done on the client in the future? Meaning in the browser/android app?

What needs to be refactored here considering this PR already targets the latest vips? 

Also is replacing intervention with vips an acceptable direction?

shleeable:
Yes, 
1. we have upgraded to the latest intervention 4.1, which I believe should allow us to do HDR in the future.
2. yes, we are talking about transcoding on the web or mobile in the future which might allow us to handle this.

Hopefully we can get the HDR supported in the future... and I will look into compiling the latest vips on the docker container for support.

but I think we'd need to wait until we support this.

jpeg-xl is not out in anything yet? https://caniuse.com/jpegxl

but avif might be a good default?

dshatz:
I may not know enough about the architecture here but what are the benefits to transcoding at the client?

I imagine it's more difficult to implement separately for each platform (how will transcoding work in js?). 
At the same time, third party clients may not transcode at all, so something like the changes in this PR would be needed anyway? 

It seems implementing it once on the backend is much simpler, as shown here, but I may be misunderstanding.

shleeable:
tbh, I think client side is more critical for the video to offload the transcoding. images can be handled locally and verified.

I am going to keep this PR open because I think my goal would be to consider AVIF as our primary image instead of JPEG... and that would allow for HDR

major problem is people using ubuntu or similar are going to be on the old vips for a long long time.

PR #6654, opened 2026-07-15
Add github workflow to build containers to ghcr.io

ilkka-ollakka:
Builds arm64 and amd64 containers and stores those into ghcr.io registry.

Builds dev and staging branches and tags

Uses docker builder that spaws matrix of jobs in native hosts in parallel build instead qemu approach.

Repository package visibility need to be checked to be public to allow people to use build containers.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,52 @@
+name: Container building for ghcr.io
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


shleeable:
Completed in another PR

ilkka-ollakka:
do note that the qemu approach is way slower than this in arm64 building. But great that the container builds are back

PR #6783, opened 2026-08-25
Use Mastodon username convention for OIDC

thomasjacumin:
I made a small change to apply the same username convention as Mastodon when creating a new user with OIDC. The reference code is on Mastodon [github](https://github.com/mastodon/mastodon/blob/main/app/models/concerns/user/omniauthable.rb), in particular the part
```ruby
def ensure_valid_username(starting_username)
    starting_username = starting_username.split('@')[0]
    temp_username = starting_username.gsub(/[^a-z0-9_]+/i, '')
    temp_username.truncate(30, omission: '')
end
```

I tested the code on my own instance.


shleeable:
Thanks for the PR. Saw this is your first touch with the project!

Do you have the ability to write a test unit for this? We're starting to build more coverage.

thomasjacumin:
Yes, I can try to write one.

thomasjacumin:
I wrote many test cases within the `RemoteOidcTest` class, inspired by the existing code.

Issue #6893, opened 2026-08-28
[Enhancement]: Location search does not support filtering by country


hohoho1886:
### Description

As discussed in #2840, when adding a location to a photo, user cannot search by country. As shown in the video, this makes it hard to find the right result when many cities share the same name across different countries.

**Proposed easy fix**: let users type a comma to separate the city and country, like "San Francisco, USA". Then filtering results so that only places matching both the city name and the country name are shown. I am aware this is not fuzzy search, which would be more ideal but requires a bigger change.

https://github.com/user-attachments/assets/53a19115-f78f-43c8-9b91-43c5d35424a9

### Use-case

Currently, city search results are ordered alphabetically, so cities with common names (e.g. "San Francisco") that exist in many countries can push the one you're looking for past the visible result limit. Letting users narrow results by country makes it possible to find and select the correct location when it would otherwise be buried.

### Acknowledgements

Yes

@shleeable I can prepare a PR for this easy fix if you give me the green light, or feel free to mark this as not planned.
Thank you so much for your time !

shleeable:
feel free to send a PR but the web UI is currently being completely replaced with a new NuxtJS frontend.

So it might not last very long :)

shleeable:
PR was merged. thank you

PR #6955, opened 2026-08-29
Add country filtering to location search

hohoho1886:
## Why this PR is needed
As mentioned in #6893, when user adds a location when posting a photo, the search engine should support search by country to be more convenient.

## What this PR does
Implement filter by country in database query. 

## Testing
I added 3 cities to the local database and tested location search using the `City, Country` format.

[Screencast from 2026-08-29 16-37-30.webm](https://github.com/user-attachments/assets/6a1a7994-978b-4e6e-802b-8f1aeaeebb7a)

## Note
I am aware that the web UI is being replaced by the new frontend. However, this PR only modifies the backend location search API. Hope it will be still compatible with upcoming frontend implementation


shleeable:
Thank you for the fix to this issue.

hohoho1886:
> Thank you for the fix to this issue.

Let me know if something needs to be changed

hohoho1886:
@shleeable Let me know if anything needs to be changed
Thanks for your time !

PR #6974, opened 2026-08-30
RFC: Mastodon-compatible Web Push subscription API

vinzgreg:
> **Status: draft / RFC.** This is not merge-ready and I'm not asking for it to be
> merged. It's here to make a case for a direction, with working code attached so
> the discussion has something concrete to argue with. I've since found real flaws
> in my own implementation — including one that defeats its main security control —
> and they're documented in full below rather than left for a reviewer to trip over.
>
> What I'd like is a read on the **concept**. If the direction is wanted, I'll do the
> work properly.

## The observation this started from

Pixelfed already carries everything Web Push needs, and never wires it up:

- `laravel-notification-channels/webpush` in `composer.json`
- `config/webpush.php`, with `VAPID_*` env config
- the `push_subscriptions` table (`2023_12_04_041631`)
- `HasPushSubscriptions` on `App\Models\User`
- the `push` OAuth scope, declared in `AppServiceProvider`
- and, in `ApiV2Controller::instance()`:

  ```php
  'configuration' => [
      'vapid' => [
          'public_key' => config('webpush.vapid.public_key'),
      ],
  ```

So `/api/v2/instance` already tells every Mastodon client that this server does Web
Push. A client reads that key, goes looking for `POST /api/v1/push/subscription`
where Mastodon's API documents it, and finds nothing there.

## Why it seems worth doing

Push currently works in one configuration: the official app, talking to an instance
whose admin has been issued a gateway key. `/api/v1.1/push/*` is gated on the
`X-PIXELFED-APP` header and delivery runs through a hosted gateway.

**Admins can't enable push themselves.** Self-hosting the whole stack still means
requesting a key for a service you don't run. Web Push needs no such thing —
generate a VAPID pair, restart the queue, and notifications go straight from your
queue worker to whatever push service your users' clients nominate.

**Third-party clients can't offer push at all.** Even where an admin has a gateway
key, it only helps the official app. A browser, a PWA, or an alternative native app
has no way in. With a standard subscription endpoint, a client that already speaks
Mastodon's Push API needs no Pixelfed-specific code. A native app, which has no Web
Push transport of its own, can point the subscription at a relay that terminates Web
Push and re-emits over APNs or FCM — the server only ever speaks RFC 8291 to a URL.

@bil-ash raised the browser-only case in the comments, which is the one this most
obviously serves and the one I'd underweighted.

## What's here

Six commits, +1397 / −0 across 18 files, each droppable from the top down:

| | |
|---|---|
| `Add the Mastodon-compatible Web Push subscription API` | endpoint, SSRF guard, migration, route |
| `Deliver Web Push for likes, follows, mentions, comments and DMs` | `WebPushNotifyPipeline` + call sites |
| `Notify over Web Push when a follow request arrives` | separable |
| `Cover the Web Push endpoint guard and configuration check` | unit tests |
| `Document the Web Push notification API` | drop if docs belong in `pixelfed/docs` |
| `Give a session-authenticated browser a subscription identity of its own` | bug fix, see below |

Delivery hooks the places that already decide a notification is warranted, alongside
the existing Expo dispatch rather than in place of it, and outside the
`NotificationAppGatewayService` gate — a Web Push subscription works on an instance
that has never configured Expo. `maybeDispatch()` is the single entry point and
honours `notify_enabled` and the per-type `notify_*` switches.

---

# Known flaws

I went looking for reasons to reject this. These are the ones I found and confirmed
by testing. The first three I'd treat as blocking.

### 1. The endpoint pinning fails open — the main security control doesn't work

The design was: re-resolve each endpoint at send time and pin the connection to the
validated addresses with `CURLOPT_RESOLVE`, closing the DNS-rebinding window. The
pin is keyed on `parse_url()`'s raw host; curl keys its resolve cache on the
**punycode** form. Where they differ, curl silently discards the pin and performs
its own lookup:

```
пример.рф              pinned->127.0.0.1   actual_ip=89.253.237.94   ok
xn--e1afmkfd.xn--p1ai  pinned->127.0.0.1   actual_ip=-               Could not connect
```

Identical host, one pinned and one not. Reaching this through the API also needs a
nameserver that answers the raw non-punycode query, so it isn't casual — but the
real problem is structural: **nothing verifies the pin applied.** IDN is one way it
silently no-ops; Guzzle falling back to its stream handler without ext-curl is
another, since that handler ignores curl options entirely.

Worse, the code asserts the opposite in its own comments — "an endpoint that cannot
be pinned is skipped rather than sent unpinned… failing closed". That is not true,
and I should not have written it without testing it. Normalising with
`idn_to_ascii()` closes the demonstrated case (and fixes legitimate IDN endpoints
being wrongly rejected), but the honest fix is to connect to the validated address
directly and carry the hostname in `Host`/SNI, so there is nothing to silently skip.

### 2. Blocking DNS in the request path is a denial-of-service

`gethostbynamel()` and `dns_get_record()` have no timeout parameter in PHP, and both
run synchronously inside `POST /api/v1/push/subscription`. A domain whose
authoritative servers blackhole queries blocks the worker for the resolver's full
retry budget, twice (A then AAAA). The only limit on the route is the global
512/min throttle, so one authenticated account can occupy every PHP-FPM worker. This
needs no sophistication at all and is the more practical of the two blockers.
Validation belongs behind a cached resolver with an explicit socket timeout, or
entirely in the queue.

### 3. `filter_var(NO_PRIV_RANGE | NO_RES_RANGE)` is not an SSRF denylist

All of these are classified public on PHP 8.4:

| Address | Reaches |
|---|---|
| `100.64.1.5`, `100.100.100.100` | CGNAT 100.64/10 — Tailscale's range; Alibaba Cloud metadata |
| `64:ff9b::a9fe:a9fe` | NAT64 well-known prefix → 169.254.169.254 |
| `2002:7f00:1::1` | 6to4 encoding of 127.0.0.1 |
| `fec0::1`, `224.0.0.1`, `2001:db8::1` | site-local, multicast, documentation range |

NAT64 and CGNAT are the two that bite in real deployments. `::ffff:127.0.0.1`,
`0.0.0.0`, octal and decimal notation are correctly rejected — and those are exactly
the cases my test file covers, so the tests demonstrate a thoroughness the code
doesn't have. This wants an explicit CIDR denylist.

### 4. Registering an endpoint can delete another user's subscription

`push_subscriptions.endpoint` is `UNIQUE`, and the vendor trait's
`updatePushSubscription()` deletes a row owned by someone else before creating
yours. The victim gets no error and simply stops receiving pushes. Endpoint URLs are
high-entropy, so this needs one to leak first — but the controller should be
returning a 409 rather than invoking that path. `pruneOldestSubscriptions()` also
does a read-then-write with no lock, so two concurrent subscribes to the same new
endpoint can race into the unique constraint and produce a 500.

### 5. Nothing ties a subscription's lifetime to its token

`access_token_id` has no FK and no cleanup. Revoking an app in settings leaves its
subscription intact, so a revoked client keeps receiving the notification stream.
Mastodon destroys the subscription with the token — I cited its per-token model as
justification for the column while omitting the half that makes it safe. Relatedly,
once a Passport token refreshes, `DELETE` matches nothing and unsubscribe silently
no-ops against an orphaned row.

### 6. Subscribing flips an account-wide preference

`notify_enabled` also gates the Expo path. A user who deliberately turned
notifications off gets them switched back on — including to the official app — by
any client that subscribes on login. It is never restored on unsubscribe. My comment
defends this on the grounds that the endpoint would otherwise be a silent no-op,
which is a real problem but not one to solve by mutating an unrelated preference.

### 7. The follow-request dedupe marker leaks

`FollowRequestObserver::deleted()` only fires for Eloquent deletes. The federated
cancel and reject paths, and both account-deletion pipelines, use query-builder mass
deletes and fire nothing — so the marker survives its full 90-day TTL and silently
swallows a later genuine follow. The docblock claims cancellation clears it; for the
federated case it doesn't. A long-lived cache key as a correctness mechanism across
uncovered deletion paths is the wrong shape regardless of the TTL.

### 8. "Mastodon-compatible" overstates it

Only `POST` and `DELETE` exist. Mastodon also has `GET` (clients call it on launch
to see whether they're subscribed) and `PUT` (how `alerts` are changed), so `alerts`
is permanently unsettable and the response echoes account-wide values a client never
set. The entity omits `follow_request`, `update`, `admin.sign_up`, `admin.report`,
and `policy` is not implemented. Boosts don't push on either transport, since
`SharePipeline` dispatches nothing. Federated DMs are announced as "New Mention",
and the local DM path isn't hooked at all.

### 9. Smaller things

`WebPushService::enabled()` returns true with only `pem_file` set, but the
subscription response then returns `server_key: null`, so the client can't subscribe.
`buildWebPushClient()` sits outside the try/catch, so a malformed VAPID key puts
every notification into `failed_jobs` with no config-time validation. Retries re-send
to subscriptions that already succeeded, and any non-2xx that isn't 404/410 is
treated as transient. Endpoints are re-resolved on every send with no caching. And
this creates a `docs/` directory that doesn't exist upstream — that's a repo-layout
question to ask before a feature PR, not inside one, so drop that commit.

---

## Testing

`staging` at `00670aa3a`, same environment both sides (Passport keys generated,
Redis available):

```
staging baseline:  546 passed, 0 failed
this branch:       573 passed, 0 failed   (+27 = the added tests)
```

`vendor/bin/pint --test` passes on 1458 files.

An earlier revision of this description reported large failure counts on both sides.
That was my test environment missing Passport keys and Redis, not anything wrong
with `staging`, and I've corrected it — apologies for the noise.

Note also that PRs targeting `staging` don't run the PHP workflows: `php-pint.yml`,
`php-larastan.yml` and `php-laravel-tests.yml` are all `pull_request: branches:
[dev]`. So none of the above was independently verified by CI, and CONTRIBUTING.md
directs contributions at `staging`. That may be worth a look separately.

The server-side implementation this is extracted from has run on a public instance
since late July 2026 against an iOS client and a Web Push → APNs relay. That is not
a claim about this branch: it's ported to `staging`'s restructured tree, and commit
six has never run anywhere.

## What I'm asking

Is this a direction you want? If yes, I'll rework it properly — connect-to-validated-IP
instead of the pin, a real CIDR denylist, DNS out of the request path, token-lifecycle
cleanup, `GET`/`PUT` and per-subscription `alerts`, and drop the docs commit. If the
answer is that push should stay first-party, that's a useful answer too and I'll stop
here rather than keep polishing something unwanted.

Happy to split any of it into smaller PRs.


bil-ash:
I am planning to host a small pixelfed server after it stabilises(hopefully within a month) with just the web ui and no mobile apps and (web )push notifications was a requirement for my use case. Thanks @vinzgreg for the PR. However, what I understood (my understanding can be wrong as well) is that the PR does not add any browser side code for actually subscribing to push notifications from browser- something like `self.addEventListener('push', event => {`  in the service worker. Would be nice if you add that as well in the PR.

vinzgreg:
@bil-ash  You're right, and thanks for flagging it — the PR is server-side only. It adds the subscription endpoint and the delivery pipeline, and nothing under `resources/` or `public/`. No service worker push listener, no subscribe call. A browser can't use it as it stands.

Your question helped me on something, I found some additional flaws of this implementaiton. It made me check how the web UI would authenticate against this endpoint, and there was a real bug: the web UI doesn't send a bearer token, so Passport authenticates the session cookie and hands back a TransientToken — which has no `id` property at all. Every browser subscription would have been stored with a null identifier, and unsubscribing in one browser would have torn down every other one. That's fixed, thank you.

Then I kept pulling on the thread and found worse things in my code. The SSRF guard's connection pinning silently fails open in some cases, DNS resolution runs unbounded inside the request path (an easy way for one account to stall a server's workers), and the address denylist misses CGNAT and NAT64 ranges. All three are written up in the description now, and I've **moved the PR to draft** — it's a 
**proposal for a direction rather than something anyone should merge**.

My intention is to raise the topic of supporting push, so any a3rd party app can build on it.

dansup:
@vinzgreg This is amazing, thank you! As [I mentioned here](https://mastodon.social/@dansup/117320011172712569) this does need some work, but can be a solid foundation for this feature. I will keep it open until it does ship!

PR #6998, opened 2026-08-31
chore: remove dead beyondcode/laravel-websockets config

alies-dev:
## Summary

- `config/websockets.php` references `BeyondCode\LaravelWebSockets\*` classes, but that package isn't in composer.json, no service provider registers it, and nothing in app/ or routes/ reads `config('websockets')`.
- Dead file left over from a removed package. Deleting it clears 14 `UndefinedClass` Psalm errors (currently baselined).

PR #6999, opened 2026-08-31
chore: bump psalm/plugin-laravel to v3.15.7, vimeo/psalm to 6.16.1

alies-dev:
Stacked on #6998, built on top of it (diff will narrow to just this PR's changes once #6998 merges; no push access to create a base branch on pixelfed/pixelfed directly, so the two branches can't share a base ref across the fork).

## Summary

- `composer update psalm/plugin-laravel vimeo/psalm -W --ignore-platform-req=php` (composer.json's `config.platform.php` pin of 8.3.0 predates 6.16.1's tightened PHP version bands; actual runtime PHP satisfies the real constraint, only the synthetic resolution floor doesn't).
- Unblocks psalm/plugin-laravel's `HasFactory` template auto-resolution (upstream fix in psalm/psalm-plugin-laravel#612), clearing all 58 `MissingTemplateParam` false positives plus the `public/index.php` `MissingFile` entry.
- Baseline regenerated against the new plugin/analyzer version.


dansup:

PR #7000, opened 2026-08-31
fix: clear easy-win Psalm findings (abort() NoValue, docblock/property fixes)

alies-dev:
Stacked on #6999 (diff will narrow once #6998 and #6999 merge; no push access to create a shared base branch on pixelfed/pixelfed across the fork, see #6999).

## Summary

- Drop the leading `return` on `abort()` calls across 8 controllers/middleware. `abort()` is declared `@return never`, so `return abort(x)` returns a value that can't exist (Psalm `NoValue`); `abort()` already halts execution, dropping `return` changes nothing at runtime.
- `ApiV1Dot1Controller`: fully qualify `@return StatusTransformer`, the unqualified name resolved against the controller's own namespace and never matched an existing class.
- `AppRegisterController` / `PersonalAccessTokenController`: use Passport's public `getToken()` instead of the magic `->token` property, avoiding the `InaccessibleProperty` flag on `PersonalAccessTokenResult`'s protected `$token`.
- `Autolink`: `@var autolinkActiveUsersOnly` named the property as its own type; it's a `bool`.
- `FixUsernames`: type-hint the empty `collect([])` so the later `foreach` variable isn't `NoValue`.
- `StatusUser`: drop `private_key`/`public_key` from `$longText`, they're already redacted and `continue`'d earlier in the loop, so keeping them here was dead overlap Psalm flagged as paradoxical.

Baseline regenerated; `StoryIndexService`'s `Redis::pipeline()` `NoValue` and `Regex.php`'s static-property reference stay baselined (tracked upstream in psalm/psalm-plugin-laravel#1421, and not worth a structural rewrite respectively).


dansup:

